PR #17, opened 2017-05-29
Support declaration:true in tsconfig.json and emit declaration files

rossipedia:
This PR modifies the plugin to respect `declaration: true` in the `tsconfig.json` file.

When enabled, this plugin will collect the generated `*.d.ts` files from the typescript compilation, and write them out after the bundle has been written.

These files will be written out to the same folder as the destination bundle (possibly make this a configurable folder?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,12 @@ Plugin takes following options:
 
 * `rollupCommonJSResolveHack`: false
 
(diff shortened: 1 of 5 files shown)


ezolenko:
I'll check it out soon, thanks!

PR #25, opened 2017-07-27
Make declarations share directory root with generated bundle(s)

wessberg:
# What is it

This PR makes the generated declaration files share the same directory root as the bundle generated by rollup.

Before, the declaration file would respect the *outDir* property of a `tsconfig.json` file and write declaration files to that directory.

This has two significant issues:
- If the *outDir* isn't the same as the generated .js files, Typescript doesn't know where to go look for the typings without manual work on the user's end.
- If more than one transform target is used in a rollup bundle (such as es2015 and umd), the declaration files will only be generated once and live outside the root of the generated folders

A previous PR has been submitted which unfortunately did not resolve this issue and, additionally, didn't handle scenarios where multiple compile targets are defined in the rollup configuration.

## What does it do

This is only relevant if the Typescript configuration has the compiler option `declaration: true`.

For the rollup configuration:
```javascript
export default {
	entry: ...,
	plugins: ...,
        sourceMap: true,
	targets: [
		{
			dest: "dist/umd/index.js",
			format: 'umd',
			moduleName: ...,
		},
		{
			dest: "dist/es2015/index.js",
			format: 'es',
		}
	]
};
```

...Which generates the destination folder structure:
```
dist
     es2015
          index.js
          index.js.map
     umd
          index.js
          index.js.map
```

...The declaration files will be rooted in the same directory:
```
dist
     es2015
          index.js
          index.js.map
          index.d.ts
     umd
          index.js
          index.js.map
          index.d.ts
```

## What it doesn't fix

The well-known issue remains where files that only export types are tree-shaken away before the `transform` hook of the plugin is ever called and thus it cannot generate declarations for those files. This is nothing new. Hopefully, we'll find a way someday.

## Other stuff

I've allowed myself to refactor a bit. Rather than import namespaces, as most modules did before, now named imports are used where possible. I've also fallen back to using the built-in `fs` module where possible rather than relying on `fs-extra` which is still used where required. I've moved some of the helper functions into separate files to make the code a bit easier to read.

I've also added more types to enforce stronger type-safety.

### DIFF
--- a/src/get-options-overrides.ts
+++ b/src/get-options-overrides.ts
@@ -1,0 +1,11 @@
+import {CompilerOptions, ModuleKind} from "typescript";
+
+export function getOptionsOverrides(): CompilerOptions {
+	return {
+		module: ModuleKind.ES2015,
+		noEmitHelpers: true,
+		importHelpers: true,
+		noResolve: false,
+		outDir: process.cwd(),
+	};
+}
--- a/src/get-options-overrides.ts
+++ b/src/get-options-overrides.ts
@@ -1,11 +1,13 @@
 import {CompilerOptions, ModuleKind} from "typescript";
+import {IOptions} from "./ioptions";
 
-export function getOptionsOverrides(): CompilerOptions {
+export function getOptionsOverrides({useTsconfigDeclarationDir}: IOptions): CompilerOptions {
 	return {
 		module: ModuleKind.ES2015,
 		noEmitHelpers: true,
 		importHelpers: true,
 		noResolve: false,
 		outDir: process.cwd(),
+		...(useTsconfigDeclarationDir ? {} : {declarationDir: process.cwd()}),
 	};
 }
--- a/src/host.ts
+++ b/src/host.ts
@@ -1,14 +1,14 @@
-import * as fs from "fs-extra";
-import * as ts from "typescript";
-import * as _ from "lodash";
+import {LanguageServiceHost as TypescriptLanguageServiceHost, IScriptSnapshot, ParsedCommandLine, ScriptSnapshot, sys, CompilerOptions, getDefaultLibFilePath} from "typescript";
+import {existsSync} from "fs";
+import {has} from "lodash";
 
-export class LanguageServiceHost implements ts.LanguageServiceHost
+export class LanguageServiceHost implements TypescriptLanguageServiceHost
 {
 	private cwd = process.cwd();
-	private snapshots: { [fileName: string]: ts.IScriptSnapshot } = {};
+	private snapshots: { [fileName: string]: IScriptSnapshot } = {};
 	private versions: { [fileName: string]: number } = {};
 
-	constructor(private parsedConfig: ts.ParsedCommandLine)
+	constructor(private parsedConfig: ParsedCommandLine)
 	{
 	}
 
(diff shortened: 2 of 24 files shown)


ezolenko:
I'll look at it in the next couple of days, thanks!

src/partial.ts
```diff
@@ -0,0 +1,3 @@
+export declare type Partial<T> = {
+	[P in keyof T]?: T[P];
```
ezolenko: Nice

src/index.ts
```diff
@@ -1,117 +1,47 @@
...
-// The injected id for helpers.
-const TSLIB = "tslib";
-let tslibSource: string;
-try
-{
-	// tslint:disable-next-line:no-string-literal no-var-requires
-	const tslibPath = require.resolve("tslib/" + require("tslib/package.json")["module"]);
-	tslibSource = fs.readFileSync(tslibPath, "utf8");
-} catch (e)
+let watchMode = false;
```
wessberg: It seems to me what you are implying is that variables defined in the global scope is accessible from outside the module they are declared within. That is incorrect. For it to persist, it must be bound to the global object (`global` in Node) and execute within the same execution environment.

So, running multiple instances would not make those variables interfere with one another.

src/get-options-overrides.ts
```diff
@@ -0,0 +1,11 @@
+import {CompilerOptions, ModuleKind} from "typescript";
+
+export function getOptionsOverrides(): CompilerOptions {
+	return {
+		module: ModuleKind.ES2015,
+		noEmitHelpers: true,
+		importHelpers: true,
+		noResolve: false,
+		outDir: process.cwd(),
```
wessberg: Nice catch in terms of declarationDir from tsconfig :-)

My expectation is for Rollup to be the decider of the placement of declaration files. The tsconfig properties that relate to output directories or filenames, in my opinion, should be overridden so that the destination that the consumers of the plugin defines in Rollup's configuration is always used. That is also the only way to make Rollup's multiple target build-step with declarations "just work" without having to maintain multiple tsconfig files with different declaration directories.

src/host.ts
```diff
@@ -1,14 +1,14 @@
-import * as fs from "fs-extra";
-import * as ts from "typescript";
-import * as _ from "lodash";
+import {LanguageServiceHost as TypescriptLanguageServiceHost, IScriptSnapshot, ParsedCommandLine, ScriptSnapshot, sys, CompilerOptions, getDefaultLibFilePath} from "typescript";
```
wessberg: I agree, syntax such as `import {Foo, Bar} as Baz from "module"` would be a nice addition to the spec

wessberg:
I've added replies to your comments. Thanks for reviewing. I'll add a revision that also handles declarationDir settings

wessberg:
But we could, as you suggest, support both. I just think the default behavior should be to follow Rollup and then optionally override that behavior to instead follow tsconfig's outDir/declarationDir :-)

wessberg:
Please see changes.
I've:
- Added handling for `declarationDir`.
- Added a configuration property: `useTsconfigDeclarationDir` to emit declaration files inside the directory given in tsconfig. By default, declarations will be placed within the same directory as the bundle generated by rollup.
- Revised README to reflect these changes.

src/get-options-overrides.ts
ezolenko: Yes, values provided in rollup should override tsconfig, but in their absence we should use tsconfig values. This will make more setups just work (and will not break existing setups).

src/index.ts
```diff
@@ -1,117 +1,48 @@
...
-const TSLIB = "tslib";
-let tslibSource: string;
-try
-{
-	// tslint:disable-next-line:no-string-literal no-var-requires
-	const tslibPath = require.resolve("tslib/" + require("tslib/package.json")["module"]);
-	tslibSource = fs.readFileSync(tslibPath, "utf8");
-} catch (e)
+// tslint:enable-next-line:no-var-requires
+let watchMode = false;
```
ezolenko: Hm, can't find my earlier comment on this... O_o 

Anyway, the reason those would fail is that those lets _are_ in global scope (check how they are transpiled). For example code like this will use single instance of those vars where it shouldn't:
```js
const rollup = require("rollup").rollup;
const ts = require("rollup-plugin-typescript2");

rollup({
    entry: "project1/main.ts",
    plugins: [
        ts(),
    ],
})
// ....

rollup({
    entry: "project2/main.ts",
    plugins: [
        ts(),
    ],
})
// ...
```

wessberg: That's right, I thought you meant that they were bound to the global object. :-) I'll move them inside the method invocation so that they are local to each invocation of the plugin's wrapper function.

ezolenko:
Thanks, I'll test it a bit and release a build in a few days.

Issue #27, opened 2017-08-04
Local TypeScript from node_modules


rvikmanis:
Is it somehow possible to use project-local `typescript`?

ezolenko:
I think it does that by default? Plugin does `import "typescript"`. If you set verbosity to high enough level (2 or 3), it will print typescript version it picked up, if you have different local and global versions, you should be able to see which one is used.

If you have a different name for typescript though...

rvikmanis:
I think it always loads the version you have specified in package.json of
your project, i.e latest minor at 2.x

ezolenko:
Ok, I added `typescript` option (in master), try requiring your version of typescript and passing that in.

rvikmanis:
Seems to work. Thanks!

PR #28, opened 2017-08-09
Move `typescript` to dev dependencies

rvikmanis:
Fixes issue #27.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "lodash": "^4.17.4",
     "object-hash": "^1.1.8",
     "rollup-pluginutils": "^2.0.1",
-    "typescript": "^2.3.4",
     "tslib": "^1.7.1",
     "resolve": "^1.3.3"
   },
@@ -53,7 +52,8 @@
     "rollup": "^0.43.0",
     "rollup-plugin-typescript2": "github:ezolenko/rollup-plugin-typescript2#master",
     "rollup-watch": "^4.0.0",
-    "tslint": "^5.4.3"
+    "tslint": "^5.4.3",
+    "typescript": "^2.3.4"
   },
   "repository": {
     "type": "git",


ezolenko:
Thanks! 

I'd rather add an option to override used typescript instead, with whatever user requires manually themselves. (I won't be able to work on it for some time though) This pr will break builds for some people. 

There was a similar issue in the original plugin, they settled on having ts as direct dependency iirc.

Could you elaborate what's your use case, maybe there is a better way?

rvikmanis:
We're using your plugin in a big project, and simply haven't yet gotten around to updating our code for typescript 2.4.

I believe this issue would be a deal breaker for many, or would fragment the rollup-typescript plugin space even more by encouraging forks.

ezolenko:
I see. I was going to suggest trying moving it to peerDependencies, but then I saw it is already there... And it is also depreciated in npm3... (https://codingwithspike.wordpress.com/2016/01/21/dealing-with-the-deprecation-of-peerdependencies-in-npm-3/)

If you are using npm3, could you try relaxing dependency rule for typescript to `2.x` and see if you end up with one typescript in npm_modules that matches your project?

wessberg:
Imagine the following scenario:

- This PR is merged.
- A user (or automated build system for continuous integration), who has no local or global installation of Typescript `npm install`'s this package.
- The build fails: Typescript could not be resolved.

This plugin *does* have a **runtime** dependency on Typescript since the Language Service is used by the plugin. Hence, making it a development dependency will not only be a breaking change for many, but also make usage of automated builds via Travis/CircleCI/etc impossible without adding a hard dependency on Typescript which not all users may be aware of.

I agree with @ezolenko that we should make it possible to pass in another version of Typescript as an argument. Typescript v>=2.4.2 would then still be installed but not conflict with the version your team is using.

It would solve the problem without breaking changes and reduce possible errors and confusion for new users

ezolenko:
Custom typescript option is in master now. If that works as expected we can close this PR.

PR #35, opened 2017-09-28
improve dependencies

lmk123:
After I upgrade to v0.6.0, it show me an error:

```
Error: Cannot find module 'rollup-pluginutils'
```

So I move `rollup-pluginutils` to _dependencies_.

BTW I add `typescript` and `rollup` to _peerDependencies_, so if user doesn't install it(or version incompatible), NPM will show a tip in console.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,12 +28,16 @@
   "dependencies": {
     "fs-extra": "^4.0.2",
     "tslib": "^1.7.1",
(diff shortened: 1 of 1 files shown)


ezolenko:
Ah, looks like I was overly enthusiastic when cleaning up dependencies. Thanks.

PR #44, opened 2017-11-23
override noEmit option to false when set

frankwallis:
Currently if you have ```noEmit: true``` in your tsconfig.json file then running rollup with this plugin just errors without telling you why it failed. Its quite common to have this option set I think to prevent lots of files being created when running tsc. It may also be worth improving the error message when nothing is emitted by the typescript compiler.


ezolenko:
Looks reasonable, can't think of any way `noEmit: true` would make sense in this context.

Thanks!

ezolenko:
Yep, I'll publish tonight or tomorrow

PR #69, opened 2018-03-23
Add 'sourceMapCallback' option to enable advanced mapping use cases

lephyrus:
**Motivation**
I am trying to do a Rollup build that transpiles my Typescript sources and specs for a Karma run, while also instrumenting the sources in a way that maps back to the original Typescript, without writing intermediate files to disk. I also want the final Rollup bundle to have correct source maps for in-browser debugging.

**Problem**
This plugin provides Rollup with the maps for the transformation from Typescript to Javascript, which correctly maps the resulting bundle to the original sources. When instrumenting the code with Istanbul as part of the process, the coverage info will not correctly apply to the original typescript, however. One approach is to use `remap-istanbul` (e.g. via `karma-remap-coverage`), but for that you need the source maps for the Typescript > Javascript transformation. These are not accessible fior a Rollup plugin, unless written to disk, which is not what this plugin does (for good reason). If I set `inlineSourceMap` to true in the Typescript compiler options, the Istanbul instrumenter correctly sees this as an "input sourcemap" and the coverage maps back to the Typescript sources perfectly. With such a configuration, `rollup-plugin-typescript2` on the other hand does not "see" the sourcemap and does not pass it back to Rollup, which leads to incorrect mappings for the generated bunde. The other way to solve this is to "manually" supply the Istanbul instrumenter with an input source map. This is what I got working.

**Solution**
With the help of the modification proposed in this PR and a custom instrumenter plugin, I got the coverage report to match the Typescript sources while also having the correct source map for the bundle produces by Rollup. 🎉 

```js
// karma.conf.js
// ...
  config.set({
    // ...
    files: [
      // ...
      { pattern: 'specs.ts', watched: false }
    ],
    preprocessors: { 'specs.ts': ['rollup'] },
    reporters: ['coverage-istanbul'],
    mime: { 'text/x-typescript': ['ts','tsx'] },
};
```
```js
// rollup.config.js
// ...
let typescriptSourceMaps = new Map();

module.exports = {
  plugins: [
    resolve(),
    typescript(
      tsconfigOverride: { include: 'specs.ts' },
      clean: true,
      sourceMapCallback: (id, map) => typescriptSourceMaps.set(id, map)
    }),
    commonjs(),
    instrument({
      include: 'src/**/*.ts',
      exclude: '**/*.spec.ts',
      getInputSourceMap: (id) => typescriptSourceMaps.get(id)
    }),
  ],
  output: {
    format: 'iife',
    sourcemap: 'inline'
  },
};
```
```js
// rollup-plugin-instrument.js
import { createInstrumenter } from 'istanbul-lib-instrument';
import { createFilter } from 'rollup-pluginutils';

export default function instrument({ include, exclude, getInputSourceMap, debug } = {}) {
  const filter = createFilter(include, exclude);
  const instrumenter = createInstrumenter({
    esModules: true,
    produceSourceMap: true
  });
  getInputSourceMap = getInputSourceMap || (() => undefined);

  return {
    name: 'instrument',

    transform(code, id) {
      if (!filter(id)) return;

      let inputMap;
      try {
        inputMap = JSON.parse(getInputSourceMap(id));
      } catch (error) {
        inputMap = undefined;
      }

      const instrumentedCode = instrumenter.instrumentSync(code, id, inputMap);
      const outputMap = instrumenter.lastSourceMap();

      return { code: instrumentedCode, map: outputMap };
    }
  };
}
```

**Notes**
* Sorry, it's a lot of explanation for a very small change. But man, this stuff was driving me crazy.
* I think omitting the `clean: true` option leads to problems with this approach, not sure why.
* Possibly related issue: https://github.com/artberri/rollup-plugin-istanbul/issues/13, 
* By the way, I could not figure out how to limit all these changes to the dist files. I think they are caused by new versions of dependencies. I don't understand how `package.lock.json` is supposed to work. I really don't. 🤷‍♂️


src/index.ts
```diff
@@ -216,6 +216,10 @@ export default function typescript(options?: Partial<IOptions>)
 				const map = _.find(output.outputFiles, (entry) => _.endsWith(entry.name, ".map"));
 				const dts = _.find(output.outputFiles, (entry) => _.endsWith(entry.name, ".d.ts"));
 
+				if (pluginOptions.sourceMapCallback && map) {
```
ezolenko: If you move this down to line 256 or so and grab map out of `result.map`, it will work with `clean: false`. Might have to save map.text in a separate property on `result` if you must have it as a text instead of parsed object.

Treat it similar to `dts` property if you need to do that.

ezolenko:
Don't worry about changes in dist, I'll rebuild it anyway.

Looks good aside from breaking on cache (see comment inline)

src/index.ts
ezolenko: You can use result.map directly if you remove JSON.parse in instrument plugin

lephyrus: * I clearly did not spend enough time to try and understand the transform function in full - makes sense now, thanks. 👍 
* There is a very good reason for handing the callback the JSON string rather than a reference to the `map` object, as I've learned the hard way: Rollup will mutate the object (by decoding the base64 VLQ mappings), so by the time the intrumenter gets hold of it, it's no longer usable.

I'll update the PR shortly.

lephyrus:
I still don't fully get how caching works, but it looks to me like the sourcemap JSON will still only be parsed once (unless a file changes) with the how it's done now.

ezolenko:
Ok, I fixed some formatting and I think there was a bug in the way I was returning maps from cache, but rollup might have been covering for it.

@lephyrus Could you check if current version works for you both clean and from cache?

lephyrus:
@ezolenko Yep, seems to work perfectly either way. Thanks for incorporating this feature so quickly!

lephyrus:
@ezolenko Can I ask if you're planning to do a release soon?

ezolenko:
Yep, most likely on the weekend sometime. I'm waiting on confirmation for one other issue, but that's not critical.

PR #123, opened 2018-12-01
Add ignore `dist` from git repo, exclude source for npm bundles.

allex:
ignore `dist` from git repo and exclude `src` of npm bundles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /.idea
 /build-self
 /build
+/dist
--- a/package.json
+++ b/package.json
@@ -7,8 +7,7 @@
   "jsnext:main": "dist/rollup-plugin-typescript2.es.js",
   "types": "dist/index.d.ts",
   "files": [
-    "dist",
-    "src"
+    "dist"
   ],
   "keywords": [
     "rollup-plugin-typescript2",


ezolenko:
Dist is in git for a reason, it allows using the plugin directly from git `"rollup-plugin-typescript2": "github:ezolenko/rollup-plugin-typescript2#master",`. This is useful for testing out bug fixes etc.

Sources in npm bundle are for source maps -- they seem to reference them by relative path. They are also included verbatim in the maps, so I'm not sure what's the correct state here is. Anybody knows if source code inside source maps sufficient or if actual files need to be nearby? Does it depend on tool support?

allex:
We should strip `dist` from git repo just because for cleanup our project commit logs for code reviews. 
And the src in npm will make local `node_modules` install bigger and bigger.   I thinks it's should be ok, just not a problem regard to source map issues.

ezolenko:
src is now dropped from npm package since 0.19.1.

Another significant pro for keeping dist is that I can review the changes there before commit to master or publish. Plugin dogfooding itself is a large part of testing. While it is possible to diff versions before publishing, it is much more in your face this way. Also leaves an audit trail of sorts.

Since most of the dependencies are bundled, I also have a chance to detect unexpected changes in those as well.

PR #127, opened 2018-12-31
Do not use deprecated rollup options

ooflorent:
Replace `ongenerate` and `onwrite` by `generateBundle` hook.


ezolenko:
Thanks

ezolenko:
Actually it doesn't work -- `this` is not an instance of the plugin. Easily fixed with saving a copy of the returned object, but this will break compatibility with rollup before 1.0.

PR #128, opened 2019-01-03
Revert "Do not use deprecated rollup options"

ezolenko:
Reverts ezolenko/rollup-plugin-typescript2#127


ooflorent:
What's the reason on this revert? What was wrong?

ezolenko:
@ooflorent I left comment on the original PR -- `this` in `generateBundle` was rollup's context object, not the object plugin created (so didn't have the methods it was expecting and was basically broken).

There is also a problem of dropping support for all rollup older than 1.0.0. I'm looking if it is worth doing that -- on one hand it is tempting to drop backwards compatibility, on the other hand this is only a warning right now, so not worth it. 

I don't see a way to both get rid of warning and support old rollup, but we _can_ support future rollup where they stop doing old calls (by both having `generateBundle` and original names for `ongenerate` and `onwrite`. This won't get rid of warnings though.

ooflorent:
Alright. `generateBundle` is also available in 0.68. I don’t know when it was introduced but maybe it is safe to migrate to it while being backward compatible ?

ezolenko:
Yeah, looks like it works in 0.68, I committed reworked version

PR #142, opened 2019-03-08
 if `file`property not exists fallback to `dir`.

cuyl:
`file` property can only be used if not more than one chunk is generated.if `file` not exists fallback to `dir`.
see: https://rollupjs.org/guide/en#big-list-of-options `output.dir` `output.file`


ezolenko:
Thanks, in 0.20.0 now

PR #151, opened 2019-05-17
Update diagnostic message according to standard tool parsing.

warrenseine:
MSBuild and many other tools / IDEs will correctly parse the message if ":" is added right after the error code.

See https://blogs.msdn.microsoft.com/msbuild/2006/11/02/msbuild-visual-studio-aware-error-messages-and-message-formats/ for reference.


warrenseine:
For reference, I'm currently sanitizing the error output with `stripAnsi(e.message).replace(/(TS[0-9]+)/, "$1:"))` where `e` is the `Error` rollup raises containing the printed diagnostic of the TypeScript compiler error.

ezolenko:
Thanks, I'll make a release soonish

PR #167, opened 2019-08-20
Move cache directory to node_modules

### LINKED ISSUE #133: Move cache directory to `node_modules`
It's common practice to use `node_modules/.cache/[namespace]` as cache directory. 

Benefits of having a cache directory in `node_modules`:

1. No need to update `.gitignore` as `node_modules` is always ignored.
2. It invisible to users.

nickmccurdy:
Fixes #133

The default `cacheRoot` is now `./node_modules/.cache` (which is the de facto standard cache location of npm packages), making it easier to use the package without needing to ignore any additional directories. The `cacheRoot` can still be changed if the user needs the original location for some reason.

## Maintenance note
I removed the cache from our own gitignore, but the version saved to `dist` still uses the old `cacheRoot`. After this is merged, it should be updated with `npm run build` and then `.rpt2_cache` can be deleted before committing. If you'd prefer, I could do it in this PR, but I wasn't sure what the process for updating `dist` was.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 /node_modules
 /npm-debug.log
 /typings
-/.rpt2_cache
 /.vscode
 /.idea
 /build-self
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ See [#108](https://github.com/ezolenko/rollup-plugin-typescript2/issues/108)
 
 	Set to true for clean build (wipes out cache on every build).
 
-* `cacheRoot`: `./node_modules/.cache`
+* `cacheRoot`: `node_modules/.cache/rollup-plugin-typescript2`
 
 	Path to cache. Defaults to a folder in node_modules.
 
--- a/README.md
+++ b/README.md
@@ -134,9 +134,9 @@ See [#108](https://github.com/ezolenko/rollup-plugin-typescript2/issues/108)
 
 	Set to true for clean build (wipes out cache on every build).
 
-* `cacheRoot`: `./.rts2_cache`
+* `cacheRoot`: `./node_modules/.cache`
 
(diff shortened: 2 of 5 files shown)


nickmccurdy:
@timdorr Thanks for the idea!

ezolenko:
Few problems: 

1. needs it's own subroot (can't just go into `.cache`, few other packages might be using same path)
2. I don't think there is always a local `./node_modules`, some projects could be using purely global node modules. 
3. If 2 is fixed, we end up with global cache and multiple unrelated builds on the same system sharing cache folder. This can open a sizable can of hookworms... Cache code must be much more robust before that can happen safely.

Default location should be the one likely to cause the fewest surprises.

nickmccurdy:
> 1. needs it's own subroot (can't just go into `.cache`, few other packages might be using same path)

The plugin already creates its own subdirectories prefixed with `rpt` so they don't conflict. This should already work with other packages, so I don't think we need another nested folder.

> 2. I don't think there is always a local `./node_modules`, some projects could be using purely global node modules.

That's pretty unlikely, though if you'd like I could add some error handling code that simply creates `node_modules` if it doesn't exist yet.

> 3. If 2 is fixed, we end up with global cache and multiple unrelated builds on the same system sharing cache folder. This can open a sizable can of hookworms... Cache code must be much more robust before that can happen safely.

We should be able to solve this by using the node_modules of the actual installation path of the plugin, rather than the current directory. For example if you're in a project with node_modules and are using the package globally, the global node_modules where the package is installed globally is used.

It seems like [find-cache-dir](https://github.com/avajs/find-cache-dir) would solve some of these issues (though I'm not sure how it treats global packages), and it's also extremely popular and well tested (over 11 million downloads this week, and used in very popular packages like ava, coveralls, and nyc, which I've used regularly with no caching issues).

ezolenko:
Looks good, thanks!

PR #170, opened 2019-09-08
Add .js extension to virtual tslib

Andarist:
Rollup has `preserveModules` option which is used to skip flat bundling and writing instead to disk a directory structure resembling the input one. This plugin makes tslib a "virtual" module by default - this means that it virtually creates an input file for Rollup to consume, but obviously it's not part of the input directory structure. To deal with it in combination with `preserveModules` Rollup creates a `_virtual` output directory and put such modules in it. As filename of such a virtual module it uses "as is" the given identifier - in this case `\0tslib`. Problem is that it doesn't add any extension to it and this trips other tools if you don't include the extension in the virtual identifier - i.e. webpack and it's resolving algorithm.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,7 +10,7 @@ import { IOptions } from "./ioptions";
 import { Partial } from "./partial";
 import { parseTsConfig } from "./parse-tsconfig";
 import { printDiagnostics } from "./print-diagnostics";
-import { TSLIB, tslibSource, tslibVersion } from "./tslib";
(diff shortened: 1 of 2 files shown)


ezolenko:
Thanks, in 0.24.1 now

mnlbox:
@Andarist, @ezolenko anyway to prevent the creation of `_tslib.js` in `preserveModules` mode?

agilgur5:
@mnlbox please see https://github.com/ezolenko/rollup-plugin-typescript2/issues/240#issuecomment-1250104318 for an answer to that question.

Issue #180, opened 2019-10-09
Why does the plugin write declaration files directly?


marijnh:
As opposed to using [`emitFile`](https://rollupjs.org/guide/en/#thisemitfileemittedfile-emittedchunk--emittedasset--string) on the rollup context?

The main issue I'm running into is that declarations aren't emitted at all when using the rollup API directly, because that calls the `generateBundle` hook with `isWrite==false`. But more generally, it would be useful to have these in the return value from `generate`, so that I can feed them into a `.d.ts` bundling tool.

marijnh:
As an additional, possibly obvious data point, if I just crudely replace the `tsModule.sys.writeFile` call with 

```javascript
cx.emitFile({
  type: "asset",
  source: entry.text,
  fileName: require("path").basename(fileName)
})
```

where `cx` is the `this` from `generateBundle`, the plugin seems to continue to correctly emit declaration files. Though I guess just passing the basename as `fileName` isn't correct when the declaration files have their own directory structure.

ezolenko:
I think this API didn't exist when the plugin was originally written (or I missed it completely). This does make a lot of sense though, feel free to make a PR. (If not, I'll fix it, eventually :)).

marijnh:
Working on a PR, but for some random reason the filenames provided for emitted assets can not be relative or absolute paths (can only emit inside of Rollup's output dir), which makes it hard to respect the `useTsconfigDeclarationDir` option (which may be anywhere). Do you have an idea on how to address this? Is directly writing the file ourselves when the file name is outside of the output dir too much of a kludge?

ezolenko:
Looks like there is a problem with emitting files during `generateBundle` -- multiple targets will cause same file being emitted multiple times and conflict.

If I track what was emitted and not emit it after first target, watch mode breaks (doesn't update types in all targets).

Try running `build`, then `build-self`, then `build-self` again to see it.

I fixed part of it in `emit` branch, but types-only files are still a problem, because those are known only at generate stage.

marijnh:
Strange. My own build process also uses multiple targets for a single bundle, and doesn't run into this error. It uses `bundle.generate` instead of `bundle.write`, but those seems to call the `generateBundle` hook in the same way.

It does seem like the bundle object is reused between calls to `generate` or `write`, which I don't really get (it would seem each write generates its own files).

[...]

PR #181, opened 2019-10-11
Use PluginContext.emitFile to emit declarations (when possible)

marijnh:
See issue #180

See [this comment](https://github.com/ezolenko/rollup-plugin-typescript2/issues/180#issuecomment-540986125) for an explanation of the logic that chooses between `writeFile` and `emitFile`.


ezolenko:
Looks good, but it is better if declarations are either all written or all emitted. So if `pluginOptions.useTsconfigDeclarationDir` is true, use `tsModule.sys.writeFile` for everything, and if not, emit everything. It would be easy to document in readme then and there will be no surprises downstream. Things should either work or not work, worst failure mode is when some files suddenly don't get processed by downstream plugins just because folder structure changed.

To resolve naming conflicts in the second case, path separators can be replaced with underscores, so `outdir/subdir/name.d.ts` becomes `outdir/subdir_name.d.ts`.

marijnh:
Always directly writing with `useTsconfigDeclarationDir` is a good idea. Implemented in the updated patch.

I don't quite get your suggestion to substitute underscores. We _can_ emit subdirectories (they just have to have the shape `foo/bar.d.ts`, not `../bar.d.ts`), and mangling file names would break imports of those files from other declaration files.

ezolenko:
Looks good, thanks. I'll make a release sometime this week hopefully.

PR #184, opened 2019-10-21
Upgrade rollup

marijnh:
With Rollup 1.25.1, the build (including build-self, multiple times) works again even with patch 551a14e773141cd993bb9d1ffd349c92481edf11 .

However, since the peerDependency still says `">=0.68.0"`, users might still run into issues when trying to run the plugin with an older Rollup that still has the bug. Should we also set the peer dep to the current version? Or consider Rollup bugs outside of the responsibility of the plugin? Or block on that?

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "lodash": "4.17.15",
     "object-hash": "1.3.1",
     "rimraf": "3.0.0",
-    "rollup": "^1.20.3 ",
+    "rollup": "^1.25.1 ",
     "rollup-plugin-commonjs": "10.1.0",
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-re": "1.0.7",


ezolenko:
Yes, we should set peer dependency to current version as well (I'll do it before releasing, in a few days hopefully, unless you want to make another quick PR). As long as minor version is different, we shouldn't break anybody who doesn't upgrade explicitly.

marijnh:
(Would really love to see a release. Let me know if I can do anything to speed that up.)

ezolenko:
Ok, two broken releases before first coffee and some coffee later, I see that types are not written at all currently with rollup 1.25.1 and 1.26.0... 

You can repro it if you update to master and do `npm build` (should see no differences in git) and then `npm build-self` -- will see all the type definitions gone (I updated build scripts to blow away dist before building, apparently npm prebuild step is not automatic...)

marijnh:
I think this is another manifestation of the same Rollup concurrency bug that caused the original issue—the declaration files are only written to the `build-self` dir, because Rollup accidentally uses the bundle object for one of the outputs for all of them in the context of the plugin hooks, so each of the `generateBundle` calls ends up writing to the same object, thus causing files to only appear in one of the output directories. Frankly, I'm somewhat surprised Rollup works as well as it does at all, given stuff like this.

I'm sorry for pulling you into this mess. I guess you should maybe roll all this stuff back for the time being and put out a non-deprecated release to get rid of the 'this package is deprecated' warning on npmjs.org. I'll continue to work with the Rollup maintainers to hopefully eventually sort out the brokenness.

ezolenko:
Ah, makes sense, no worries. I just unpublished last two versions, should get rid of the warnings (should have done that yesterday, I thought you couldn't delete packages from npm)

marijnh:
Rollup 1.26.3 no longer has the issue that caused the declaration files to disappear here. (Several build-self runs leave the output unchanged.) So, yeah, if you're still interested in this, it might make sense to try again.

ezolenko:
Sure, could you create another PR?

marijnh:
See #190

PR #190, opened 2019-11-04
Restore emitFile behavior, upgrade Rollup

marijnh:
This upgrades Rollup to 1.26.3, which doesn't have the problem that blocked #184, and restores patch 551a14e773141c, which with that Rollup version no longer breaks the self-build.

If you want the tree shape to look differently, let me know.


ezolenko:
Yep, looks good, both with useTsconfigDeclarationDir on and off.

Hopefully it won't break in next rollup release :)

marijnh:
Supposedly Lukas is working on a more thorough fix for the underlying issues, so with luck that means that there won't be more nonsense like this.

PR #195, opened 2019-11-25
Fix syntax error in transformer example

chocolateboy:


ezolenko:
This is valid syntax either way?

chocolateboy:
[It's not](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Returning_object_literals). Try it:

  $ cat temp.js

```javascript
const transformer = (service) =>
{
     before: [ keysTransformer(service.getProgram()) ],
     after: []
};
```

  $ node temp.js

```
temp.js:4
     after: []
          ^
SyntaxError: Unexpected token ':'
```

ezolenko:
Ah, it's a function returning object literal. Missed that part, thanks.

PR #203, opened 2020-02-11
(fix): upgrade object-hash to support async/await syntax

### LINKED ISSUE #105: Does not work with plugin which containing async/await syntax
This plugin does not work with plugin which containing async/await syntax caused by [object-hash issue](https://github.com/puleos/object-hash/pull/68) and [tscache.ts](https://github.com/ezolenko/rollup-plugin-typescript2/blob/707c6fc300ce5b6ed4d11992b8e466f328fa88b2/src/tscache.ts#L105+L111). I think this is hard to fix object-hash because there is no way to detect asyncfunction now. So is there any alternative without object-hash?

#### Environment 
- Related
  - https://github.com/smooth-code/svgr/pull/147
  - https://github.com/smooth-code/svgr/blob/master/packages/rollup/src/index.js#L11

#### Versions
- typescript: 2.8.3
- rollup: 2.1.1
- rollup-plugin-typescript2: 0.14.0

#### rollup.config.js
```js
import svgr from '@svgr/rollup';
import typescript from 'rollup-plugin-typescript2';

export default {
  ...
  plugins: [
    replace({ 'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV) }),
    svgr(),
    typescript({
    })
  ],
  ...
};
```

#### tsconfig.json
No relevant.

#### package.json
No relevant.

```
[!] (rpt2 plugin) Error: Unknown object type "asyncfunction"
src/components/atoms/Icon/index.ts
Error: Unknown object type "asyncfunction"
    at Object._object (/Users/vwxyutarooo/Projects/kouzoh/mercari-web-jp-component/node_modules/rollup-plugin-typescript2/node_modules/object-hash/index.js:218:17)
    at Object._function (/Users/vwxyutarooo/Projects/kouzoh/mercari-web-jp-component/node_modules/rollup-plugin-typescript2/node_modules/object-hash/index.js:319:14)
    at Object.dispatch (/Users/vwxyutarooo/Projects/kouzoh/mercari-web-jp-component/node_modules/rollup-plugin-typescript2/node_modules/object-hash/index.js:185:30)
    at /Users/vwxyutarooo/Projects/kouzoh/mercari-web-jp-component/node_modules/rollup-plugin-typescript2/node_modules/object-hash/index.js:246:18
    at Array.forEach (<anonymous>)
    at Object._object (/Users/vwxyutarooo/Projects/kouzoh/mercari-web-jp-component/node_modules/rollup-plugin-typescript2/node_modules/object-hash/index.js:242:21)
    at Object.dispatch (/Users/vwxyutarooo/Projects/kouzoh/mercari-web-jp-component/node_modules/rollup-plugin-typescript2/node_modules/object-hash/index.js:185:30)
    at /Users/vwxyutarooo/Projects/kouzoh/mercari-web-jp-component/node_modules/rollup-plugin-typescript2/node_modules/object-hash/index.js:260:23
    at Array.forEach (<anonymous>)
    at Object._array (/Users/vwxyutarooo/Projects/kouzoh/mercari-web-jp-component/node_modules/rollup-plugin-typescript2/node_modules/object-hash/index.js:259:20)
```

agilgur5:
- object-hash v2.0.2 adds/fixes support for async functions
  - object-hash v2 has no breaking changes, just drops testing of
    Node < 6
    - Node 5 was EOL April 2018, and on top of that Node 6
      was EOL April 2019, and Node 8 was EOL December 2019, so this is
      still more tested support than is necessary

- objectHashIgnoreUnknownHack isn't useful for its original purpose to
  workaround this, but is still useful for other types of
  objects/syntaxes that object-hash may not handle yet
  (e.g. async generator functions), so leave the hack in
  - might be good to add a warning though? so that those who used this
    option for async/await can upgrade and remove it (and have a
    stable cache again)

<hr>

**Note**: I added an initial commit in here that auto-updates the `package-lock.json` as it hadn't been updated since at least v0.25.2, and I didn't want that to be part of the diff for the actual `object-hash` upgrade (which is a very tiny diff).

I also noticed that `package.json` is set to v0.25.4, but v0.25.4 was never released.

<hr>

Fixes #105 with no need for a hack.

This was just fixed in https://github.com/puleos/object-hash/pull/90 , which is a very similar fix to the one originally referenced in #105 (https://github.com/puleos/object-hash/pull/68).

Like I mentioned in https://github.com/puleos/object-hash/pull/90 , we're currently facing this problem in [TSDX](https://github.com/jaredpalmer/tsdx) in several issues: https://github.com/jaredpalmer/tsdx/issues/294 , https://github.com/jaredpalmer/tsdx/issues/358#issuecomment-566232991 , https://github.com/jaredpalmer/tsdx/issues/278#issuecomment-563184800 , https://github.com/jaredpalmer/tsdx/issues/496#issuecomment-584026269, and a few in https://github.com/jaredpalmer/tsdx/issues/379 , so it's one of the top issues.

The current `objectHashIgnoreUnknownHack` either makes the cache stale or with `clean` makes for no caching, both of which are very suboptimal workarounds. Would be great to actually fix the main problem `objectHashIgnoreUnknownHack` addresses (it may still be used for future syntax, other object types, etc) instead of having hacky workarounds for the most common case of async functions!

The bundling of dependencies (#80, #70) also means there's no way for TSDX to internally change `rollup-plugin-typescript2`'s dependencies. This means we don't really have a way of fixing this without waiting for for a new release from this library, which is also suboptimal IMO.

This issue was also one of the main reasons behind TSDX temporarily switching to [`@wessberg/rollup-plugin-ts`](https://github.com/wessberg/rollup-plugin-ts) in https://github.com/jaredpalmer/tsdx/issues/200

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "rollup-plugin-typescript2",
-  "version": "0.25.2",
+  "version": "0.25.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -10,7 +10,7 @@
       "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.5.0"
+        "@babel/highlight": "^7.0.0"
       }
     },
     "@babel/highlight": {
@@ -19,9 +19,9 @@
       "integrity": "sha512-7dV4eu9gBxoM0dAnj/BCFDW9LFU0zvTrkq0ugM7pnHEgguOEeOz1so2ZghEdzviYzQEED0r4EAgpsBChKy1TRQ==",
       "dev": true,
       "requires": {
-        "chalk": "2.4.2",
-        "esutils": "2.0.3",
-        "js-tokens": "4.0.0"
+        "chalk": "^2.0.0",
+        "esutils": "^2.0.2",
+        "js-tokens": "^4.0.0"
       }
     },
     "@types/colors": {
@@ -30,7 +30,7 @@
       "integrity": "sha512-7jNkpfN2lVO07nJ1RWzyMnNhH/I5N9iWuMPx9pedptxJ4MODf8rRV0lbJi6RakQ4sKQk231Fw4e2W9n3D7gZ3w==",
       "dev": true,
       "requires": {
-        "colors": "1.3.3"
+        "colors": "*"
       }
     },
     "@types/estree": {
@@ -51,7 +51,7 @@
       "integrity": "sha512-bCtL5v9zdbQW86yexOlXWTEGvLNqWxMFyi7gQA7Gcthbezr2cPSOb8SkESVKA937QD5cIwOFLDFt0MQoXOEr9Q==",
       "dev": true,
       "requires": {
-        "@types/node": "8.0.47"
+        "@types/node": "*"
       }
     },
     "@types/graphlib": {
@@ -78,7 +78,7 @@
       "integrity": "sha512-il4NIe4jTx4lfhkKaksmmGHw5EsVkO8sHWkpJHM9m59r1dtsVadLSrJqdE8zU74NENDAsR3oLIOlooRAXlPLNA==",
       "dev": true,
       "requires": {
-        "@types/node": "8.0.47"
+        "@types/node": "*"
       }
     },
     "@types/resolve": {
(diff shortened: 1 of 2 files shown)


ezolenko:
Thanks, I'll check it out

ezolenko:
> I also noticed that package.json is set to v0.25.4, but v0.25.4 was never released.

This is to support pointing npm directly to github, committed version is always higher than last released.

ezolenko:
In 0.26.0 now

agilgur5:
Big thanks for the super fast turnaround @ezolenko !! Hopefully this ends a lot of hackiness 😄

PR #219, opened 2020-03-26
Use imports with @rollup/pluginutils

NotWoods:
Switches to the newest pluginutils version and uses an import statement rather than a require statement.


ezolenko:
In 0.27.0 now, thanks.

PR #135, opened 2019-02-02
Add unit tests

brekk:
Hey @ezolenko,

I saw in the [README](https://github.com/ezolenko/rollup-plugin-typescript2#contributing) that you needed some unit tests -- I tried to add some decent coverage for the files here (though I think I may have misunderstood the build process (specifically the `tsProxy` stuff); if this branch is useful I'm happy to amend / continue this branch but as I'm about to start a new job in a couple days I figured it would be better to post this as is rather than let it languish.

screen shot 2019-02-02 at 1 58 16 pm

Additionally this is the first time I've written unit tests for TS -- I did some dirty checking of instances via `instance as any` but I think in general it has decent coverage for the files I started tests for.

I hope it's ok, feedback appreciated.

Thanks,

Brekk

PR #221, opened 2020-03-29
(fix): declaration maps should have correct sources

### LINKED ISSUE #204: declaration maps (*.d.ts.map) have incorrect relative paths in `sources` unless `declarationDir` is set

This comes from downstream TSDX at https://github.com/jaredpalmer/tsdx/issues/479 (and https://github.com/jaredpalmer/tsdx/issues/135), which I fixed in https://github.com/jaredpalmer/tsdx/issues/488 by adding a `declarationDir`, but I'm not sure if this isn't just a bug here. I believe the potential bug is either in https://github.com/ezolenko/rollup-plugin-typescript2/blob/ac821249f38a083c12071df8b420bea8c527d76e/src/get-options-overrides.ts#L31 or in https://github.com/ezolenko/rollup-plugin-typescript2/blob/df241da73f355140ae5d38b748e22fac3be89b65/src/index.ts#L364 but I'm not sure. Also not sure if this is related to #136 or not.

The tl;dr on those related issues is that TSDX takes `./src/index.ts` as input and outputs `./dist/index.js`, `./dist/index.d.ts`, and if `declarationMap` is `true`, `./dist/index.d.ts.map`, which has a `sources` attribute.
- The problem is that when `declarationDir` is _not_ set, `sources` will _incorrectly_ point to `index.ts`
- When `declarationDir` _is_ set to `./dist`, `sources` will _correctly_ point to `../src/index.ts`.
  - (with `useTsconfigDeclarationDir: true` ofc)

But in both cases, whether `declarationDir` is set to `./dist` or not, everything gets output to `./dist` via Rollup config.

If I'm understanding the code correctly, this is because the TS Language Service isn't creating the files when `declarationDir` isn't set, and so the relative paths are generated incorrectly before `emitFile` is called.
So that _seems_ like a bug here, but maybe I'm missing something or its not fixable without setting `declarationDir`, sorry if that's the case 😅 

#### Environment 

[TSDX @ master](https://github.com/jaredpalmer/tsdx/tree/68f26c90d81aa65ff46098b59d6f5ae9cf591cbd)

#### Versions
- typescript: 3.7.5
- rollup: 1.27.8
- rollup-plugin-typescript2: 0.26.0

#### rollup.config.js

The only difference is what's listed above (`useTsconfigDeclarationDir` and `declarationDir`), but current, non-trivial config is [here](https://github.com/jaredpalmer/tsdx/blob/68f26c90d81aa65ff46098b59d6f5ae9cf591cbd/src/createRollupConfig.ts)

#### tsconfig.json

`declarationDir` is the only relevant one but here's [a fixture](https://github.com/jaredpalmer/tsdx/blob/68f26c90d81aa65ff46098b59d6f5ae9cf591cbd/test/fixtures/build-default/tsconfig.json)

#### package.json

Probably not relevant but [here](https://github.com/jaredpalmer/tsdx/blob/68f26c90d81aa65ff46098b59d6f5ae9cf591cbd/package.json)

No errors, so I don't think it's relevant? Can run this on a fixture if needed.

agilgur5:
- change declarationDir default to be Rollup's output destination
  instead of cwd
  - previously, it was set to cwd if useTsconfigDeclarationDir wasn't
    true, however, declarations aren't output to cwd, but to Rollup's
    output destination, so this was incorrect

- also change the one line in the docs that says it'll be
  `process.cwd()`; every other reference says it'll be the output dest

Fixes #204 and downstream https://github.com/jaredpalmer/tsdx/issues/479 . Better solution to downstream https://github.com/jaredpalmer/tsdx/issues/488

Couldn't change `get-options-overrides` because the `options` hook only gets `InputOptions`; it's not till `generateBundle` that we get access to `OutputOptions`.

RE: multiple output dirs: Rollup will call `generateBundle` for each output as far as I understand. `InputOptions` is passed to `rollup` which generates a `bundle` object, and `bundle.write` is called with each `OutputOptions` -- it's a [two step+ process](https://github.com/rollup/rollup/issues/3422#issuecomment-595706449). I think that's also why the type is `OutputOptions` and not `OutputOptions[]` (though the code doesn't use Rollup's `Plugin` interface; but that also defines it as `generateBundle` with just `OutputOptions`).

Would be great to add tests to ensure the declaration maps are correct now, but there doesn't seem to be any test set-up 😕 #135 seems to partially address that but was never completed.
Can see https://github.com/jaredpalmer/tsdx/pull/488 for how I tested `sources` there

### DIFF
--- a/src/get-options-overrides.ts
+++ b/src/get-options-overrides.ts
@@ -6,7 +6,7 @@ import * as _ from "lodash";
 import { join } from "path";
 import { IContext } from "./context";
 
-export function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot, cwd }: IOptions, preParsedTsconfig?: tsTypes.ParsedCommandLine): tsTypes.CompilerOptions
+export function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot }: IOptions, preParsedTsconfig?: tsTypes.ParsedCommandLine): tsTypes.CompilerOptions
 {
 	const overrides: tsTypes.CompilerOptions = {
 		noEmitHelpers: false,
@@ -24,11 +24,9 @@ export function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot, cwd 
 		if (preParsedTsconfig.options.module === undefined)
 			overrides.module = tsModule.ModuleKind.ES2015;
 
-		const declaration = preParsedTsconfig.options.declaration;
-		if (!declaration)
+		// only set declarationDir if useTsconfigDeclarationDir is enabled
+		if (!useTsconfigDeclarationDir)
 			overrides.declarationDir = undefined;
-		if (declaration && !useTsconfigDeclarationDir)
-			overrides.declarationDir = cwd;
 
 		// unsetting sourceRoot if sourceMap is not enabled (in case original tsconfig had inlineSourceMap set that is being unset and would cause TS5051)
 		const sourceMap = preParsedTsconfig.options.sourceMap;
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,11 +12,11 @@ import { parseTsConfig } from "./parse-tsconfig";
 import { printDiagnostics } from "./print-diagnostics";
 import { TSLIB, TSLIB_VIRTUAL, tslibSource, tslibVersion } from "./tslib";
 import { blue, red, yellow, green } from "colors/safe";
-import { relative } from "path";
+import { relative, dirname, resolve as pathResolve } from "path";
 import { normalize } from "./normalize";
 import { satisfies } from "semver";
 import findCacheDir from "find-cache-dir";
-import { PluginImpl, PluginContext, InputOptions, OutputOptions, MinimalPluginContext, TransformResult } from "rollup";
+import { PluginImpl, PluginContext, InputOptions, OutputOptions, MinimalPluginContext, TransformResult, SourceMap } from "rollup";
 import { createFilter } from "./get-options-overrides";
 
 type RPT2Options = Partial<IOptions>;
@@ -365,11 +365,27 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 				}
 				else
 				{
-					const relativePath = relative(pluginOptions.cwd, fileName);
+					// don't mutate the entry because generateBundle gets called multiple times
+					let entryText = entry.text
+					const declarationDir = (_output.file ? dirname(_output.file) : _output.dir) as string;
+					const cachePlaceholder = `${pluginOptions.cacheRoot}/placeholder`
+
+					// modify declaration map sources to correct relative path
+					if (extension === ".d.ts.map") {
+						const parsedText = JSON.parse(entryText) as SourceMap;
+						// invert back to absolute, then make relative to declarationDir
+						parsedText.sources = parsedText.sources.map(source => {
+							const absolutePath = pathResolve(cachePlaceholder, source);
+							return relative(declarationDir, absolutePath);
+						});
+						entryText = JSON.stringify(parsedText);
+					}
+
+					const relativePath = relative(cachePlaceholder, fileName);
 					context.debug(() => `${blue("emitting declarations")} for '${key}' to '${relativePath}'`);
 					this.emitFile({
 						type: "asset",
-						source: entry.text,
+						source: entryText,
 						fileName: relativePath,
 					});
 				}
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ The plugin inherits all compiler options and file lists from your `tsconfig.json
 * `noEmit`: false
 * `inlineSourceMap`: false (see [#71](https://github.com/ezolenko/rollup-plugin-typescript2/issues/71))
 * `outDir`: `./placeholder` in cache root, see [83](https://github.com/ezolenko/rollup-plugin-typescript2/issues/83) and [Microsoft/TypeScript/issues/24715](https://github.com/Microsoft/TypeScript/issues/24715)
-* `declarationDir`: `process.cwd()` (*only if `useTsconfigDeclarationDir` is false in the plugin options*)
+* `declarationDir`: Rollup's `output.file` or `output.dir` (*only if `useTsconfigDeclarationDir` is false in the plugin options*)
 * `moduleResolution`: `node` (*`classic` is [deprecated](https://www.typescriptlang.org/docs/handbook/module-resolution.html). It also breaks this plugin, see [#12](https://github.com/ezolenko/rollup-plugin-typescript2/issues/12) and [#14](https://github.com/ezolenko/rollup-plugin-typescript2/issues/14)*)
 * `allowNonTsExtensions`: true to let other plugins on the chain generate typescript, update plugin's include filter to pick them up (see [#111](https://github.com/ezolenko/rollup-plugin-typescript2/issues/111))
 


ezolenko:
Sorry for ignoring this for a while -- this breaks something -- if you build the plugin itself with this change, it drops types into `dist\node_modules\.cache\rollup-plugin-typescript2\placeholder` instead of `dist` like rollup config says.

To test do `npm build` (check changes in git) and then do `npm build-self` (check git again and you'll see lots of files moved).

agilgur5:
So I had to learn a lot about Rollup plugins and lifecycle to write this PR. Then without further guidance here had to do a lot of learning of this codebase and the TS API. As a result, it took me a while to understand how to fix this.

### Debugging the PR

> it drops types into `dist\node_modules\.cache\rollup-plugin-typescript2\placeholder` instead of `dist` like rollup config says.

I figured out why this was happening relatively quickly, that's because `outDir` is set to `${cacheRoot}/placeholder`, and by default `declarationDir` is equivalent to `outDir`. That's on this line: https://github.com/ezolenko/rollup-plugin-typescript2/blob/2f51110a8cad186dffed8aea2834bd82c44209e5/src/get-options-overrides.ts#L17

I added some logging during output that displays this as well. While `sources` is correct now, the `entry` was output to the wrong place, `${cacheRoot}/placeholder`:

```bash
entry name:  /Users/agilgur5/Desktop/GitHub/rollup-plugin-typescript2/node_modules/.cache/rollup-plugin-typescript2/placeholder/partial.d.ts.map
dtsmap entry text:  {"version":3,"file":"partial.d.ts","sourceRoot":"","sources":["../../../../src/partial.ts"],"names":[],"mappings":"AAAA,MAAM,CAAC,OAAO,MAAM,OAAO,CAAC,CAAC,IAAI;KAAG,CAAC,IAAI,MAAM,CAAC,CAAC,CAAC,EAAE,CAAC,CAAC,CAAC,CAAC;CAAG,CAAC"}
```

That can't be changed to Rollup's `output.dir` for the same reason `declarationDir` can't be; per my opening comment "it's not till `generateBundle` that we get access to `OutputOptions`."

#### Current PR doesn't work

In the current PR, the changes apparently have no effect on the output because they change `declarationDir` during `generateBundle`, but the declarations were already created during `transform`: https://github.com/ezolenko/rollup-plugin-typescript2/blob/2f51110a8cad186dffed8aea2834bd82c44209e5/src/index.ts#L261 

#### Trying different solutions

Between March - June-ish, I came back to this a few times to add more logging and understand this as well as to try to somehow set `declarationDir` properly, but failed a lot because this isn't possible basically. 

The `LanguageServer`, which has the `tsconfig` `declarationDir` setting, is created during `options`: https://github.com/ezolenko/rollup-plugin-typescript2/blob/2f51110a8cad186dffed8aea2834bd82c44209e5/src/index.ts#L114

I thought to lazily initialize this, but that's not possible, it's needed in the next plugin lifecycle steps, especially `transform`.

It can't be done in `generateBundle` where I was trying in the current PR -- it would have to be done in or before `transform` -- but `OutputOptions` like `output.dir` aren't passed until `generateBundle`. That causes quite a dilemma. 

But Rollup is designed that way intentionally, any output transformations should be left until `generateBundle`.

### Solution

Basically the only option is to transform the declaration maps' `sources` to the correct relative path to `output.dir` during `generateBundle`. I didn't want to resort to doing so as it's one more transform and seemingly in the wrong lifecycle step, but that seems to be how it's designed. It's also a bit expensive performance-wise as each declaration map JSON has to be parsed and then stringified, and sometimes the full source (in this case declaration source) is inside those, making that more perf-heavy.

I checked `@wessburg/rollup-plugin-ts` to see if it was handled similarly there, and it [indeed seems to be](https://github.com/wessberg/rollup-plugin-ts/blob/980e1dad284a7fc3cd3c2ec6c862be13099843ad/src/service/emit/declaration/bundle-declarations-for-chunk.ts#L18-L19) (it also bundles declarations, so the logic is a bit different, but still relevant):

```typescript
map = JSON.parse(text) as SourceMap;
map.file = options.declarationPaths.fileName;
```

So I'm going to rework this PR to have a similar approach and transform `sources` during `generateBundle`. Should hopefully have that done soon

src/get-options-overrides.ts
```diff
@@ -24,12 +24,6 @@ export function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot, cwd
 		if (preParsedTsconfig.options.module === undefined)
 			overrides.module = tsModule.ModuleKind.ES2015;
 
-		const declaration = preParsedTsconfig.options.declaration;
-		if (!declaration)
-			overrides.declarationDir = undefined;
-		if (declaration && !useTsconfigDeclarationDir)
-			overrides.declarationDir = cwd;
-
```
agilgur5, line 32: I removed this whole thing because `undefined` is the default ...but I think this should actually be:

```ts
        if (!useTsconfigDeclarationDir)	
			overrides.declarationDir = undefined;
```

So that it isn't set to `declarationDir` otherwise.

src/index.ts
```diff
@@ -365,11 +365,28 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 				}
 				else
 				{
-					const relativePath = relative(pluginOptions.cwd, fileName);
+					// don't mutate the entry because generateBundle gets called multiple times
+					let entryText = entry.text
```
agilgur5, line 369: this took me a few hours to debug and was by far the most time-consuming piece 😕  I originally changed `entry.text` itself which then got changed on top of itself for each output...

Ideally I think `generateBundle` should have some caching so this isn't repeated per output. I could do that with a cache key inside of `declarations{}`, but there's a good question as to what that cache key should be (in this case, I think the `declarationDir` variable is the only thing that is subject to change,  but that sounds brittle if more stuff gets added).

src/index.ts
```diff
@@ -365,11 +365,28 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
...
+
+					// modify declaration map sources to correct relative path
+					if (extension === ".d.ts.map") {
+						const parsedText = JSON.parse(entryText) as SourceMap;
+						// invert back to absolute, then make relative to declarationDir
+						parsedText.sources = parsedText.sources.map(source => {
+							const absolutePath = pathResolve(cachePlaceholder, source);
+							return relative(declarationDir, absolutePath);
+						});
+						console.log('new sources: ', parsedText.sources)
```
agilgur5, line 381: woops this was for debugging, should be removed:
```suggestion
```

agilgur5:
@ezolenko this should be good to go now -- if you run `npm run build && npm run build-self` it will change the declaration maps' `sources` to be correct now, and the files should all still be in the right places.

ezolenko:
@agilgur5 looks good now. The only problem I see is sources in type maps are in windows format if I build it on windows: `"..\\src\\context.ts"`. Any idea if this will confuse existing tooling?

agilgur5:
Ah, I don't run on a Windows machine, but if that's not normal behavior (wouldn't know), then could use `path.normalize`

ezolenko:
@agilgur5 Ok, PR is in master, could you doublecheck and I'll do a release

agilgur5:
Looks like you added `normalize` in front of the `relative` call so looks correct to me. I saw you added `normalize` in front of the `emitFile` `fileName` mapping to however -- that didn't have `normalize` in a previous version, I only changed the path, not how it was mutated: https://github.com/ezolenko/rollup-plugin-typescript2/commit/ec0568ba2c8e206372f94164e697b1469bf3f33d#diff-f41e9d04a45c83f3b6f6e630f10117feL368 . I don't know if it _should_ be normalized, but that is different from what it was before.

On `install` + `build` + `build-self`, the `package-lock.json` auto-updates to the new version and the `cjs.js.map` and `es.js.map` get updated for some reason (not possible to read the diff to understand since it's a one-liner). Not sure what that means.

agilgur5:
@ezolenko any update on the above or a release?

agilgur5:
@ezolenko you merged some of my other PRs, but didn't comment on the above and there still hasn't been a release 😕

ezolenko:
Sorry, was distracted.

When you run npm install it will pull patch updated of deps and I that might affect source maps a bit (maybe line number changed because of comments or something).

I'll do a release later today.

agilgur5:
What about this part?

> I saw you added `normalize` in front of the `emitFile` `fileName` mapping to however -- that didn't have `normalize` in a previous version, I only changed the path, not how it was mutated: [ec0568b#diff-f41e9d04a45c83f3b6f6e630f10117feL368](https://github.com/ezolenko/rollup-plugin-typescript2/commit/ec0568ba2c8e206372f94164e697b1469bf3f33d#diff-f41e9d04a45c83f3b6f6e630f10117feL368) . I don't know if it _should_ be normalized, but that is different from what it was before.

ezolenko:
Re normalize for emit file, seems to have no effects either way, rollup handles it fine, but might affect downstream plugins that want to modify maps. Normalize is more correct in principle though, so I guess we should keep it and see if problems show up.

agilgur5:
Ok, yea I read in some Rollup issues (can't find which one it was) that it allows whichever internally, which means plugins have to handle and do the normalization. https://github.com/rollup/rollup/issues/2952 suggests this is particularly relevant for `resolveId`

PR #227, opened 2020-05-07
Update tslib

ajlende:
This updates tslib to `1.11.2` so that when tslib is bundled, it doesn't carry the Apache-2.0 license with it anymore which solves a [licensing issue](https://github.com/WordPress/gutenberg/pull/21024#issuecomment-606708806) I've been dealing with.

The only change in [`tslib@1.11.2`](https://github.com/microsoft/tslib/releases/tag/1.11.2) is the new license.

> This release relicenses tslib to the 0BSD license which should allow projects that embed or bundle tslib to omit its license header.

Thank you for taking a look at this! 🙂


ezolenko:
In 0.27.1 now, thanks.

Issue #237, opened 2020-08-07
Monorepo example


desmap:
#### Environment 

I am trying to setup a monorepo with local packages. I am using pnpm workspaces but could also go with tsconfig's paths features. Whatever, I don't get it to work  and would love to see an example where somebody gets a monorepo working with this plugin.

My problem is that I get this all the time:
```
index.ts → index.js...
[!] Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)
../greeter/index.ts (2:6)
1: export default class {
2:   name: string
         ^
3:
```
note that the imported local package (`greeter`) has a index.js, index.mjs and index.d.ts file

#### Versions
- typescript: 3.9.7
- rollup: 2.23
- rollup-plugin-typescript2: 3.1.1

#### rollup.config.js
```
import alias from '@rollup/plugin-alias'
import typescript from 'rollup-plugin-typescript2'

module.exports = {
  input: 'index.ts',
  output: {
    file: 'index.js',
    format: 'module',
  },
  plugins: [
    alias({
      entries: [{ find: 'greeter', replacement: '../greeter' }],
    }),
    typescript(),
  ],
}
```

#### tsconfig.json
```
{
  "compilerOptions": {
    "target": "esnext",
    "moduleResolution": "node"
  }
}
```

#### package.json
```
{
  "name": "app",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "type": "module",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "typescript": "^3.9.7"
  },
  "dependencies": {
    "@rollup/plugin-alias": "^3.1.1",
    "greeter": "workspace:^1.0.0",
    "rollup-plugin-typescript2": "^0.27.1",
    "tslib": "^2.0.0"
  }
}
```
#### plugin output with verbosity 3

```
rpt2: built-in options overrides: {
    "noEmitHelpers": false,
    "importHelpers": true,
    "noResolve": false,
    "noEmit": false,
    "inlineSourceMap": false,
    "outDir": "~/test/app/node_modules/.cache/rollup-plugin-typescript2/placeholder",
    "moduleResolution": 2,
    "allowNonTsExtensions": true,
    "module": 5
}
rpt2: parsed tsconfig: {
    "options": {
        "target": 99,
        "moduleResolution": 2,
[...]

PR #249, opened 2020-10-24
fix: use compilerOptions.rootDir to filter files

NicolasThierion:
When compilerOptions.rootDir is specified, it is used instead of process.cwd() to filter files. 

Possible fix for #237


ezolenko:
In 0.29.0 now

PR #251, opened 2020-11-11
Fix duplicate output with multiple entry points

antross:
On Windows the normalized paths in resolveId end up in POSIX format.
This cause rollup to treat the returned path as a new piece of content.
This in turn results in duplicate output for references across entry points.

Fixed by normalizing the path to use host OS separators before returning.

See https://github.com/antross/rollup-plugin-typescript2-repro for a repro project.


ezolenko:
Looks like this is a workaround for https://github.com/rollup/rollup/issues/3094. I feel we'll have to revisit it again when something changes :)

githoniel:
when will this release?

this will also fix `preserveModuleRoot` in windows.

`preserveModuleRoot` will use path segment separation depending  by OS in Rollup but this this plugin will always use `/`

ezolenko:
@githoniel released in 0.30.0, sorry for the delay

PR #253, opened 2020-11-18
Add tslib to installation instructions

brandon-leapyear:
Version 2 of yarn uses Plug'n'Play, which means that the user's installed `typescript` (which is used because `typescript` is set as a peer dependency for this plugin) won't have access to the `tslib` marked as a dependency for this plugin.

https://github.com/yarnpkg/berry/issues/2140

People using version 2 of yarn have to manually install `tslib` in `devDependencies`. This PR makes it so that everyone would need to manually install `tslib` as well, which is what should actually happen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,9 @@ This version is somewhat slower than original, but it will print out typescript 
 ```bash
 # with npm
 npm install rollup-plugin-typescript2 typescript --save-dev
-# with yarn
+# with yarn berry
+yarn add rollup-plugin-typescript2 typescript tslib --dev
+# with classic yarn
 yarn add rollup-plugin-typescript2 typescript --dev
 ```
 
--- a/README.md
+++ b/README.md
@@ -16,11 +16,9 @@ This version is somewhat slower than original, but it will print out typescript 
 
 ```bash
 # with npm
-npm install rollup-plugin-typescript2 typescript --save-dev
-# with yarn berry
+npm install rollup-plugin-typescript2 typescript tslib --save-dev
+# with yarn
 yarn add rollup-plugin-typescript2 typescript tslib --dev
-# with classic yarn
-yarn add rollup-plugin-typescript2 typescript --dev
 ```
 
 ## Usage
--- a/package.json
+++ b/package.json
@@ -32,12 +32,12 @@
     "@rollup/pluginutils": "^3.1.0",
     "find-cache-dir": "^3.3.1",
     "fs-extra": "8.1.0",
-    "resolve": "1.17.0"
+    "resolve": "1.17.0",
+    "tslib": "2.0.1"
   },
   "peerDependencies": {
     "rollup": ">=1.26.3",
-    "typescript": ">=2.4.0",
-    "tslib": ">=2.0.0"
+    "typescript": ">=2.4.0"
   },
   "devDependencies": {
     "@types/colors": "1.2.1",
--- a/package.json
+++ b/package.json
@@ -32,12 +32,12 @@
     "@rollup/pluginutils": "^3.1.0",
     "find-cache-dir": "^3.3.1",
     "fs-extra": "8.1.0",
-    "resolve": "1.17.0",
-    "tslib": "2.0.1"
+    "resolve": "1.17.0"
   },
   "peerDependencies": {
     "rollup": ">=1.26.3",
-    "typescript": ">=2.4.0"
+    "typescript": ">=2.4.0",
+    "tslib": ">=2.0.0"
   },
   "devDependencies": {
     "@types/colors": "1.2.1",


merceyz:
This isn't Yarn PnP specific, it's required for all package managers, PnP just enforces it

ezolenko:
What happens currently? Does typescript pull its own version of tslib and plugin pulls in its own?

brandon-leapyear:
Currently, because `rollup-plugin-typescript` has `tslib` as a `dependency`, it installs it `node_modules`, and normal npm usage would find it, because npm module resolution just looks for the existence of a `node_modules` directory and sees if `node_modules/tslib/package.json` exists.

However, this is actually relying on incorrect behavior on npm's part, and yarn v2 fixes this by using a scheme called Plug'n'Play, where it doesn't use `node_modules` at all and instead explicitly exposes packages only if the packages should be exposed to the running process. If a user doesn't explicitly install `tslib`, Yarn v2 will error saying that the `rollup` build process can't find `tslib`.

Because `rollup-plugin-typescript2` doesn't actually use `tslib` (i.e. there's no `require('tslib')` or `import ... from 'tslib'` in `rollup-plugin-typescript2`), it shouldn't be a dependency, but actually a `peerDependency`, which indicates that `rollup-plugin-typescript2` doesn't directly use it but it's transitively required at runtime (when using the `typescript` library)

ezolenko:
Hm, the plugin actually uses tslib at runtime here https://github.com/ezolenko/rollup-plugin-typescript2/blob/8d0d49c063a317ad23e9d74a234db2f9f6076157/src/tslib.ts#L12

Will that cause problems? Should we add tslib to both peer dependencies and dependencies?

brandon-leapyear:
Interesting! Ok, then let's move it back to dependencies. I think, maybe, the `typescript` library should list `tslib` as a `peerDependency`, since the `typescript` library expects `tslib` to already be installed (but it's not listed in `typescript`'s dependencies)

@merceyz do you have any thoughts on this?

merceyz:
> What happens currently? Does typescript pull its own version of tslib and plugin pulls in its own?

When you enable `importHelpers` TypeScript checks your project for `tslib` but since you're not adding it as a dependency PnP is rejecting access to it and TypeScript throws. 

To illustrate this problem without PnP imagine for a moment there is no hoisting and you haven't added `tslib` as a dependency to your project, your node_modules layout will look like this:

```sh
package.json
src
  -> index.ts
node_modules
  -> typescript
  -> rollup
  -> rollup-plugin-typescript2
     -> node_modules
        -> tslib
```

Now when TypeScript has to import a helper from `tslib` and add it to `src/index.ts` it will check if it can find it from that file, which it can't.

> @merceyz do you have any thoughts on this?

The solution here is for the user to add `tslib` to their project, how that is communicated is up to @ezolenko

brandon-leapyear:
@merceyz Thanks! My question is more, is the "correct" solution to get `typescript` to add `tslib` as a `dependency/`peerDependency`?

ezolenko:
@merceyz

>Now when TypeScript has to import a helper from tslib and add it to src/index.ts it will check if it can find it from that file, which it can't.

Hm, this shouldn't be a problem unless typescript is used directly. Plugin resolves tslib itself, so when typescript tries to import it, it will generate import statement in js code (because of `importHelpers: true` override in tsconfig), plugin will resolve that to its internal copy that it loaded in `src/tslib.ts` and rollup will inject it into the bundle correctly. 

Does it break somehow?

merceyz:
> @merceyz Thanks! My question is more, is the "correct" solution to get typescript to add tslib as a dependency/peerDependency`?

The correct solution is for the user to add `tslib` to their dependencies, adding it as dependency to typescript wouldn't help (ref https://github.com/ezolenko/rollup-plugin-typescript2/pull/253#issuecomment-731402746) and if it was a peer dependency it would be optional so you wouldn't see a warning until it failed at build time anyways.

> Hm, this shouldn't be a problem unless typescript is used directly. Plugin resolves tslib itself, so when typescript tries to import it, it will generate import statement in js code (because of importHelpers: true override in tsconfig), plugin will resolve that to its internal copy that it loaded in src/tslib.ts and rollup will inject it into the bundle correctly.
>
> Does it break somehow?

As I said, TypeScript will check if it's installed before injecting it, so the plugin changing it later doesn't matter.

brandon-leapyear:
> When you enable `importHelpers` TypeScript checks your project for `tslib`

Would this be solved if TypeScript's check for `tslib` did the equivalent of `require.resolve('tslib')`?

merceyz:
That would still make you responsible for adding it as a dependency, it's not whether or not typescript can find it, it's if your project can.

brandon-leapyear:
What do you mean? Isn't this error coming about when `typescript` is compiling a `.ts` file and needs to (effectively) copy and paste `tslib` into the compiled JS? Or are you saying that `typescript` just injects an `import "tslib"` statement and separately checks to see if that would fail when executed?

ezolenko:
I think he is saying typescript itself will try to find tslib by its own means before generating import statement when transpiling ts. So itworks if ts either doesn't check and just generate import (plugin will handle it), or it checks and finds it.

ezolenko:
PR is just a readme change, I can merge it unless somebody things something else needs to be done?

PR #293, opened 2021-12-14
chore: should not lock deps version

bluelovers:


ezolenko:
Some of the reasons for rolling dependencies statically apply to locking them too :) (see discussion here: https://github.com/ezolenko/rollup-plugin-typescript2/pull/80)

ezolenko:
@bluelovers does this need a release?

bluelovers:
@ezolenko no need publish now

PR #296, opened 2022-02-01
Add `trace` method to the `LanguageServiceHost` to enable usage with `traceResolution`

Andarist:
At the moment I'm debugging a weird difference in behavior between `tsc` and `tsdx` (where the latter is using this plugin) and being able to look through the output of `traceResolution` is super helpful for me - but I had to patch this plugin locally to make it work. It would be great if this option would just be respected by the plugin.


ezolenko:
In 0.31.2 now, thanks

PR #307, opened 2022-03-30
docs: Change `@rollup/plugin-typescript` link

dev-itsheng:
I noticed that the previous link did not exist anymore, after a search it was moved to a new location, so I changed the link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 Rollup plugin for typescript with compiler errors.
 
-This is a rewrite of original rollup-plugin-typescript, starting and borrowing from [this fork](https://github.com/alexlur/rollup-plugin-typescript).
+This is a rewrite of original rollup-plugin-typescript, starting and borrowing from [this fork](https://github.com/rollup/plugins/tree/master/packages/typescript).
 
 This version is somewhat slower than original, but it will print out typescript syntactic and semantic diagnostic messages (the main reason for using typescript after all).
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 Rollup plugin for typescript with compiler errors.
 
-This is a rewrite of original rollup-plugin-typescript, starting and borrowing from [this fork](https://github.com/rollup/plugins/tree/master/packages/typescript).
+This is a rewrite of original [`rollup-plugin-typescript`](https://github.com/rollup/rollup-plugin-typescript)(The library is no longer maintained, and now available at [`@rollup/plugin-typescript`](https://github.com/rollup/plugins/tree/master/packages/typescript)), starting and borrowing from a fork by `alexlur`.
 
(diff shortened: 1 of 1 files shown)


ezolenko:
Thanks, this change makes the text not quite correct though -- this repo was started from a specific fork that was later deleted by its author (and the source repo was moved to new location as you found). That's why the link doesn't work. 

Maybe change to something like this:

```
This is a rewrite of original [rollup-plugin-typescript](https://github.com/rollup/plugins/tree/master/packages/typescript), starting and borrowing from a fork by `alexlur`.
```

dev-itsheng:
That is, its roadmap looks like this:

```text
A rollup-plugin-typescript(rollup/rollup-plugin-typescript) ----> D @rollup/plugin-typescript(rollup/olugins)
   ↓
B rollup-plugin-typescript(alexlur/rollup-plugin-typescript)
   ↓
C rollup-plugin-typescript2(this repo)
```

A is still there, but its GitHub README is declared read-only and no longer maintained, and needs to be migrated to D, and B has been removed.

I thought about it carefully and decided to clarify the relationship between them and change it to what it is now.

README.md
```diff
@@ -6,7 +6,7 @@
 
 Rollup plugin for typescript with compiler errors.
 
-This is a rewrite of original rollup-plugin-typescript, starting and borrowing from [this fork](https://github.com/alexlur/rollup-plugin-typescript).
+This is a rewrite of original [`rollup-plugin-typescript`](https://github.com/rollup/rollup-plugin-typescript)(The library is no longer maintained, and now available at [`@rollup/plugin-typescript`](https://github.com/rollup/plugins/tree/master/packages/typescript)), starting and borrowing from a fork by `alexlur`.
```
agilgur5, line 9: IMO, if this were to be merged:
- I think there's no need to link to `@rollup/plugin-typescript` since the archived repo does that already and most folks familiar with Rollup would know that too
- I don't think we should remove the link to alexur's fork, even if it doesn't exist anymore. It might be private right now or reuploaded later. Or it could be available on the Wayback Machine etc. So the link is useful even if it currently 404s

It might also be good to link to an _older_ commit of `rollup-plugin-typescript` for posterity. It looked like [`0.8.1`](https://github.com/rollup/rollup-plugin-typescript/tree/v0.8.1) may be where the fork started, or potentially earlier.
Definitely _not_ later though, as the next tag, `1.0.0`, came after this repo's first few commits, e.g. afc02fc2f92567748eef2f06c393d96589202eeb is a good reference

So, with that said, my personal suggestion would be:

```diff
This is a rewrite of the [original `rollup-plugin-typescript`](https://github.com/rollup/rollup-plugin-typescript/tree/v0.8.1), starting and borrowing from [@alexur's fork]((https://github.com/alexlur/rollup-plugin-typescript).
```

This would be simpler to read and maintains some of OP's changes as well

agilgur5: Added this in #380

Issue #286, opened 2021-10-20
node v17: Error loading `tslib` helper library


aladdin-add:
#### What happens and why it is wrong

```bash
Error loading `tslib` helper library.
/Users/weiran/repo/dian/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:25177
    throw e;
    ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /Users/weiran/repo/dian/tsdx/node_modules/rollup-plugin-typescript2/node_modules/tslib/package.json
```

##### Versions

```text
  Node: 17.0.0
  OS: macOS 12.0
  yarn: 3.0.2

   rollup: ^2.57.0 => 2.58.0 
    rollup-plugin-typescript2: ^0.30.0 => 0.30.0 
    typescript: ^4.4.3 => 4.4.4
```

#### rollup.config.js

<details>
  <summary>`rollup.config.js`: </summary>

</details>

#### tsconfig.json

<details>
  <summary>`tsconfig.json`: </summary>

</details>

#### package.json

<details>
  <summary>`package.json`: </summary>

</details>

#### plugin output with verbosity 3

<details>
  <summary>plugin output with verbosity 3: </summary>

</details>

aladdin-add:
this is from https://github.com/weiran-zsd/dts-cli/issues/78.

yoeunes:
Hello @aladdin-add did you find any solution for this error ? I got it also with` node 17`

```
  Node: v17.0.1
  npm: 8.1.0
  OS: Ubuntu
  rollup: 2.58.0
rollup-plugin-typescript2: 0.30.0
```

xiCO2k:
Checking in on this, any updates?

bluelovers:
> Checking in on this, any updates?

i think my pr can fix this

bluelovers:
can try this before merge apply

```
"rollup-plugin-typescript2": "bluelovers/rollup-plugin-typescript2#pr/node-v17-tslib-002",
```

ezolenko:
@bluelovers I'm having trouble building with node 17 (unrelated to your PR), because something breaks semver either at runtime (before upgrading commonjs plugin) or at build time, so I had to disable version check. Do you see those problems at all? 

[...]

PR #321, opened 2022-05-08
test: add initial unit test suite

agilgur5:
## Summary

Adds an initial batch of unit tests, _heavily_ updated and refactored from #135 (and several new tests added on top as well). Also adds tests to the CI process.
- Closes #135

## Description

- deps: add `jest` for testing, `ts-jest` for processing TS files, and `@jest/globals` for importing from Jest within tests
  - add a `jest.config.js` with necessary configurations
    - ignore `fixtures/` dir, only match `.spec.ts` files for tests (no test helper files)
  - `gitignore` the `coverage` directory
  
- test: adds unit tests into `__tests__` dir
  - Adds unit tests for `check-tsconfig`, `context`, `diagnostics-format-host`, `get-options-override`, `host`, `nocache`, `rollingcache`, and `rollupcontext`
    - 100% coverage for all but `get-options-overrides`; couldn't get `expandIncludeWithDirs` to work, see [in-line comment](https://github.com/ezolenko/rollup-plugin-typescript2/pull/321#discussion_r867520059)
  - Missing tests for `print-diagnostics`, `tscache`, and `index`
    - `index` really still needs heavy integration testing with the [Rollup API](https://rollupjs.org/guide/en/#javascript-api)
    - No need for tests for interface files -- could add type tests if you really wanted to though
    - `tslib` and `tsproxy` are workaround/hacky files, but may be able to test them too
      - **EDIT**: `tsproxy` is now tested indirectly as it's needed in a few of the specs

- test: uses `__tests__/__temp/` dir for generated test files
  - this is removed after testing and also `.gitignore`'d

- lint: adds `__tests__` dir to lint dirs
- ci: adds `npm run test:coverage` to `steps`
- ci: add Node 14, 16, 18, and macOS, Windows to matrix
  - drop Node 10 as latest Jest can't run on it
    - meant to do these OS/Node CI changes in a separate, independent PR, but added them here due to the Jest issues
  - this is to test newer versions of Node (e.g. #286) and OS differences like [POSIX separators](https://github.com/ezolenko/rollup-plugin-typescript2/issues?q=label%3A%22topic%3A+OS+separators%22+), case-insensitivity, etc

- docs: adds a testing section to `CONTRIBUTING.md`
  - and generally adds sections to split up "Developing"

## Review Notes

- Could use a `jest.config.ts` if we add `ts-node` as a dep. I thought you may not want to add that as a dep so I left it out for now

## Next Steps

1. Can add Coverage Reporting and a badge to the README with something like [codecov](https://codecov.io). I use this in [my own libraries](https://app.codecov.io/gh/agilgur5), but this requires manual set-up so I didn't add it in the code yet. I might have permissions to do it myself now, but not sure; also not sure if you'd want it on my account vs. yours etc
    - Can integrate with [CI](https://docs.codecov.com/docs/codecov-uploader), add a [coverage threshold check](https://docs.codecov.com/docs/commit-status) and add [a badge](https://docs.codecov.com/docs/status-badges)
    
1. Still need plenty of integration testing for the plugin itself / `index`. Should use the [Rollup API](https://rollupjs.org/guide/en/#javascript-api) for these
    - Adding common errors and edge cases (e.g. type-only imports, monorepos, etc) as regression tests would be great in particular
    - In general, this would make repros a lot easier and within the test system
1. Can optimize the above as well as existing filesystem tests by using an in-memory filesystem / `tmpfs`, e.g. [`memfs`](https://github.com/streamich/memfs) or with a [RAMDisk](https://github.com/shellscape/webpack-plugin-ramdisk/issues/3)

1. Improve coverage to 100% of already tested files
1. Add unit tests for `print-diagnostics`, `tscache`, maybe others


agilgur5:
Ack, this merge conflicts with at least one of my recently merged PRs: #319 . I'll fix that and add running tests to the `CONTRIBUTING.md` from #313 in a bit

__tests__/get-options-overrides.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+	const stubbedContext = makeStubbedContext({});
+	const filter = createFilter(stubbedContext, config, preParsedTsConfig);
+
+	expect(filter("src/test.ts")).toBe(true);
+	expect(filter("src/test.js")).toBe(false);
+	expect(filter("src/test.d.ts")).toBe(false);
+});
+
+// not totally sure why this is failing
+test.skip("createFilter -- rootDirs", () => {
```
agilgur5, line 146: So as far as I understand, I believe I'm testing the `expandIncludeWithDirs` function correctly.

And according to the original PR that wrote the code, #38, this looks like the intended usage too.
(I wrote the test before looking at that PR, fwiw)

src/check-tsconfig.ts
```diff
@@ -1,10 +1,8 @@
...
 import * as tsTypes from "typescript";
-
 export function checkTsConfig(parsedConfig: tsTypes.ParsedCommandLine): void
 {
 	const module = parsedConfig.options.module!;
 
-	if (module !== tsModule.ModuleKind.ES2015 && module !== tsModule.ModuleKind.ESNext && module !== tsModule.ModuleKind.ES2020)
-		throw new Error(`Incompatible tsconfig option. Module resolves to '${tsModule.ModuleKind[module]}'. This is incompatible with rollup, please use 'module: "ES2015"' or 'module: "ESNext"'.`);
+	if (module !== tsTypes.ModuleKind.ES2015 && module !== tsTypes.ModuleKind.ESNext && module !== tsTypes.ModuleKind.ES2020)
+		throw new Error(`Incompatible tsconfig option. Module resolves to '${tsTypes.ModuleKind[module]}'. This is incompatible with rollup, please use 'module: "ES2015"' or 'module: "ESNext"'.`);
```
agilgur5, line 7: all the `./tsproxy` -> `typescript` changes were from the original PR #135. I didn't touch them for now, just updated them.

Sounds like I may want to undo them per https://github.com/ezolenko/rollup-plugin-typescript2/pull/135#issuecomment-488047504 though

agilgur5: Fixed this up. Undid the source changes from the original author and indeed needed to use `setTypescriptModule` and then everything started to work. Builds no longer have strange output as well.

Only weird thing was that importing and using `tsModule` from `tsproxy` didn't work -- it came out as `undefined` for some reason (possibly due to ESM semantics of static exports?). But it's using the same TS as the direct import, so I just used the direct import.

__tests__/context.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+	proxy.error(() => "ftest2");
+	expect(console.log).toHaveBeenLastCalledWith("=>ftest2");
+
+	proxy.info(() => "ftest3");
+	expect(console.log).toHaveBeenLastCalledWith("=>ftest3");
+
+	proxy.debug(() => "ftest4");
+	expect(console.log).toHaveBeenLastCalledWith("=>ftest4");
+
+	expect((proxy as any).prefix).toEqual("=>");
```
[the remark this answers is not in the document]
agilgur5: I ended up refactoring out all of these and properly using the public interfaces, so resolving this now.

(or just removing some of these, such as this in case, where the behavior was _already_ tested via the public interface)

__tests__/host.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+	expect(host.readDirectory(testDir)).toEqual([testFile]);
+	expect(host.readFile(nonExistent)).toBeFalsy();
+	expect(host.readFile(testFile)).toEqual(unaryFunc);
+	expect(host.useCaseSensitiveFileNames()).toBe(process.platform === "linux");
+
+	// test misc functionality
+	expect(host.getCompilationSettings()).toEqual(testOpts);
+	expect(host.getDefaultLibFileName({})).toEqual(
+		require.resolve("typescript/lib/lib.d.ts")
+	);
```
agilgur5, line 67: Ok I tried figuring this out for a few hours (while working on a bunch of other stuff too -- some more refactors here and more PRs coming!) and it seems that `tsModule.getDefaultLibFilePath({})` just returns a wacky half-normalized path on Windows???

```
    Expected: "D:\\a\\rollup-plugin-typescript2\\rollup-plugin-typescript2\\node_modules\\typescript\\lib\\lib.d.ts"
    Received: "D:\\a\\rollup-plugin-typescript2\\rollup-plugin-typescript2\\node_modules\\typescript\\lib/lib.d.ts"
 ```
 
 Screenshot from the Actions logs (which ran a bunch of times on this...):
 
Screen Shot 2022-05-10 at 1 08 29 AM

**Comment by agilgur5** [comment]
Ok this is 99.99% done now, rebased with latest `master`, `CONTRIBUTING.md` improved, more (significant) refactors added in, more coverage added in, all tests fixed (except for one on Windows...), CI upgraded, etc

Two in-line comments remaining. One just adds more coverage if we can fix it, another is a really confusing Windows-specific path error that seems to be within TS??

File: __tests__/host.spec.ts (hunk 4 above)
Comment (agilgur5) [reply __tests__/host.spec.ts:67]: Oh... Looked at the TS source code and this indeed appears to be a bug [in TS](https://github.com/microsoft/TypeScript/blob/8f56f6b49d2fb282ff95df64f62dab2947004e95/src/services/services.ts#L2881)... was not expecting that 😅  
Thought either I was doing something wrong or the `host.ts` file had a bug in for the longest time.... the code is like 7 years old too, very surprised this hasn't come up before?? (and no existing issues on it either)

Gonna submit a PR to TS tomorrow and put a workaround in here then too

Comment (ezolenko) [reply __tests__/host.spec.ts:67]: With the amount of path normalization bugs out there every lib should basically assume everything is possible :)

**Review by ezolenko — APPROVED** [review]
Looks great! Hopefully can serve as a good platform for integration tests too eventually.

**Comment by ezolenko** [comment]
I think you should be able to merge this yourself when you are ready

File: __tests__/host.spec.ts (hunk 4 above)
Comment (agilgur5) [reply __tests__/host.spec.ts:67]: You'd think it would've been discovered in 7 years though in such a popular tool like TS itself! Guess not even TS is safe 😅 

Filed issue https://github.com/microsoft/TypeScript/issues/49050 and PR https://github.com/microsoft/TypeScript/pull/49051, will write up a tiny workaround for this till that gets merged, released, etc in the meantime

File: __tests__/get-options-overrides.spec.ts (hunk 1 above)
Comment (agilgur5) [reply __tests__/get-options-overrides.spec.ts:150]: @ezolenko do you have any idea why this test fails from a quick glance? would love to get it working; don't think I can blame TS internals for this one 😆

File: __tests__/host.spec.ts (hunk 4 above)
Comment (agilgur5) [reply __tests__/host.spec.ts]: Just checked `endsWith("lib.d.ts")` as a workaround for now -- left a TODO comment referencing this issue for future reference for a better test.

**Review by agilgur5 — COMMENTED** [review]
Think this is finally good to go! 🎉 

Time for a gigantic squash merge now... 😅

File: __tests__/get-options-overrides.spec.ts [hunk 5]
```diff
@@ -0,0 +1,151 @@
...
+
+const forcedOptions: ts.CompilerOptions = {
+	allowNonTsExtensions: true,
+	importHelpers: true,
+	inlineSourceMap: false,
+	moduleResolution: ts.ModuleResolutionKind.NodeJs,
+	noEmit: false,
+	noEmitHelpers: false,
+	noResolve: false,
+	outDir: local(`${cacheDir}/placeholder`),
```
agilgur5, line 50: Oh of course I found another Windows right before merging...

```
    -   "outDir": "D:\\a\\rollup-plugin-typescript2\\rollup-plugin-typescript2\\__tests__\\__temp\\get-options-overrides\\placeholder",
    +   "outDir": "D:\\a\\rollup-plugin-typescript2\\rollup-plugin-typescript2\\__tests__\\__temp\\get-options-overrides/placeholder",
```

This is because `get-options-overrides.ts` doesn't do a normalization when interpolating `/placeholder`, so this is an actual bug in _this_ codebase.

As this PR is just tests (no source changes) and about to be merged, just going to leave a `// TODO: ` comment here to fix this in the source in a future PR.
That may or may not be breaking? I'm surprised that some of this code still runs fine on Windows despite bugs like these... I'm guessing there might be some more redundant normalization done elsewhere which is why it doesn't throw an error?

agilgur5:
All tests passed finally and merged!

> Hopefully can serve as a good platform for integration tests too eventually.

Yea at least the test harness is built-out, just need to add test structure for integration tests.
Fortunately, I created TSDX's integration tests (https://github.com/jaredpalmer/tsdx/pull/627) and those are much more complex than the ones this plugin will require, so I hope to get a basic structure out for those soon too!

__tests__/get-options-overrides.spec.ts
ezolenko: Looks like this has nothing to do with `rootDirs`, rollup filter just doesn't like folder prefixes. This test fails as well:

```test("createFilter -- rootDirs 2", () => {
	const config = {
		...defaultConfig,
		include : ["src/*.ts+(|x)"],
	};
	const preParsedTsConfig = {
		...defaultPreParsedTsConfig,
	};

	const stubbedContext = makeStubbedContext({});
	const filter = createFilter(stubbedContext, config, preParsedTsConfig);

	expect(filter("src/test.ts")).toBe(true);
});```

ezolenko: `include : ["src/test.ts"],` also fails, but `include : ["**/src/test.ts"],` succeeds. `include : ["/src/test.ts"] [...] filter("/src/test.ts")` also works.

This likely has something to do with the way `rootDirs` are presented in typescript after parsing tsconfig -- I can't see this value in `parsedConfig.options.rootDirs` at runtime (even though it is there in types and in tsconfig), but I see it in `parsedConfig.wildcardDirectories` as an absolute path.

I think the bug is that we are using `parsedConfig.options.rootDirs` instead of `parsedConfig.wildcardDirectories`. If we switch, rollup will get absolute paths in includes and things would work.

ezolenko: I wonder if typescript deprecated `CompilerOptions.rootDirs` value or stopped writing it sometime after this functionality was implemented.

agilgur5: So [`rootDirs`](https://www.typescriptlang.org/tsconfig#rootDirs) is definitely still used by TS; I remember in https://github.com/jaredpalmer/tsdx/issues/668 people were using `rootDirs` as a workaround with Storybook, which is only 2 years old.

I did see that `@rollup/pluginutils` has had some [breaking changes in v4.0.0](https://github.com/rollup/plugins/blob/master/packages/pluginutils/CHANGELOG.md#v400) to `createFilter`, but I don't think that impacts our usage here... (It is perhaps different than what the author of #38 originally intended though?)

It seems like the problem with these tests is that `filter` actually generally takes an _absolute_ path. The usage of `filter` in the codebase seems to match that too -- only absolute paths are used with it.
The [docs](https://github.com/rollup/plugins/tree/master/packages/pluginutils#createfilter) for `pluginutils` say that `createFilter` resolves the `include`/`exclude` paths with `process.cwd()` (when `**` is not given as of v4, since that means all dirs), and I do see that in [the code](https://github.com/rollup/plugins/blob/556837f0b7e7323f237370e36394569abc2b1ef9/packages/pluginutils/src/createFilter.ts#L15).

So I actually think what's needed to get these tests to pass is just to concat the path with `process.cwd()` so it has an absolute path equivalent

agilgur5: Got it working in #329

__tests__/get-options-overrides.spec.ts
agilgur5: Fixed this in #331

agilgur5:
> Can add Coverage Reporting and a badge to the README with something like [codecov](https://codecov.io). I use this in [my own libraries](https://app.codecov.io/gh/agilgur5), but this requires manual set-up so I didn't add it in the code yet. I might have permissions to do it myself now, but not sure; also not sure if you'd want it on my account vs. yours etc
> 
> * Can integrate with [CI](https://docs.codecov.com/docs/codecov-uploader), add a [coverage threshold check](https://docs.codecov.com/docs/commit-status) and add [a badge](https://docs.codecov.com/docs/status-badges)

It doesn't seem like I have permissions to add Coverage Reporting myself; think you have to be repository "owner" or otherwise have access to the "Settings" panel of the repo, which I don't have access to.

PR #323, opened 2022-05-10
optim(ci): use `npm ci` for install and cache `npm`

agilgur5:
## Summary

Optimize CI by using `npm ci` and caching `npm` installs

## Details

- `npm ci` is installation for, well, CI, and is a good bit faster
  - c.f. https://docs.npmjs.com/cli/v8/commands/npm-ci

- cache `npm` installation with `setup-node` for speedier installs
  - upgrade `setup-node` as this was released in [v2.2.0](https://github.com/actions/setup-node/releases/tag/v2.2.0)

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -22,10 +22,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


agilgur5:
Hard to tell from just a single CI run but looks ~25% faster (basically installation is twice as fast). And there's no cache on this run just yet!

PR #324, opened 2022-05-11
Update @yarn-tool/resolve-package from 1.0.40 to 1.0.47

felipecrs:
In order to fix this issue:

https://github.com/bluelovers/ws-iconv/issues/5

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1394,31 +1394,6 @@
         "@sinonjs/commons": "^1.7.0"
       }
     },
-    "node_modules/@ts-type/package-dts": {
-      "version": "1.0.56",
-      "license": "ISC",
-      "dependencies": {
-        "@types/semver": "^7.3.9",
-        "ts-type": "^2.1.2"
-      }
-    },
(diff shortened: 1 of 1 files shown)


agilgur5:
rpt2 [doesn't pin a version of this dep](https://github.com/ezolenko/rollup-plugin-typescript2/blob/327574e8437a68a3372bb68c5105d7592d180f2c/package.json#L36), just the major, so you should be able to update this yourself in your own `package-lock` without needing rpt2 to make any changes (that is one of the benefits of not pinning, after all)

agilgur5:
Just a relevant FYI: I actually removed `@yarn-tool/resolve-package` as a dep in #326, so once a new version of rpt2 is released, you should be able to upgrade to it to remove the dep (and the transitive `@types/node` dep from there)

agilgur5:
[0.32.0](https://github.com/ezolenko/rollup-plugin-typescript2/releases/tag/0.32.0) no longer has `@yarn-tool/resolve-package` as a dep

PR #326, opened 2022-05-14
deps: upgrade `tslib` to `^2.4.0`, remove `@yarn-tool/resolve-package`

### LINKED ISSUE #286: node v17: Error loading `tslib` helper library

agilgur5:
## Summary

Resolves #286 in a simpler way by upgrading `tslib` instead of introducing any new deps
- Replaces #289 / #290 

Implements the approach I proposed in https://github.com/ezolenko/rollup-plugin-typescript2/issues/286#issuecomment-1120270016 

## Details

- `tslib` `2.4.0` is forward and backward-compatible with older and newer Node exports mechanisms, so the Node 17 error should no longer be present
  - see its [`package.json`](https://github.com/microsoft/tslib/blob/a7129c7bd500ce378ec19234247f3d0b1635e89d/package.json#L35) after https://github.com/microsoft/tslib/pull/135
    - it has the older `./` and the newer `./*` in its package exports, which should allow for `package.json` to be read in both older and newer implementations

- this allows us to remove the extra dep on `@yarn-tool/resolve-package` as well
  - other than less unnecessary deps being good, `@yarn-tool/resolve-package` is also a not well-documented package with very few users, which does not make for a good security posture for rpt2 (which has historically prioritized supply chain security in other issues around deps (c.f. #80, https://github.com/ezolenko/rollup-plugin-typescript2/pull/293#issuecomment-993918512) or, in particular, its consumers, which there are very many of (in contrast with `@yarn-tool`)
  - per my issue comment https://github.com/ezolenko/rollup-plugin-typescript2/issues/286#issuecomment-1120270016, we could also have avoided the extra dep prior to the `tslib` upgrade by resolving to absolute paths, as Node only does a ["weak" encapsulation](https://nodejs.org/api/packages.html#main-entry-point-export) of relative imports

- test: add a small unit test for `tslib.ts` to ensure that this method works and passes on different Node versions in CI
  - more a smoke test that it runs at all, the testing is additional and a bit duplicative of the source tbh

### DIFF
--- a/__tests__/tslib.spec.ts
+++ b/__tests__/tslib.spec.ts
@@ -1,0 +1,11 @@
+import { test, expect } from "@jest/globals";
+import * as fs from "fs-extra";
+
+import { tslibVersion, tslibSource } from "../src/tslib";
+
+test("tslib", async () => {
+  expect(tslibVersion).toEqual(require("tslib/package.json").version);
+
+  const tslibES6 = await fs.readFile(require.resolve("tslib/tslib.es6.js"), "utf8");
+  expect(tslibSource).toEqual(tslibES6);
+});
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,11 +10,10 @@
       "license": "MIT",
       "dependencies": {
         "@rollup/pluginutils": "^4.1.2",
-        "@yarn-tool/resolve-package": "^1.0.40",
         "find-cache-dir": "^3.3.2",
         "fs-extra": "^10.0.0",
         "resolve": "^1.20.0",
-        "tslib": "^2.3.1"
+        "tslib": "^2.4.0"
       },
       "devDependencies": {
         "@jest/globals": "^28.0.3",
@@ -1396,6 +1395,7 @@
     },
     "node_modules/@ts-type/package-dts": {
       "version": "1.0.56",
+      "dev": true,
       "license": "ISC",
       "dependencies": {
         "@types/semver": "^7.3.9",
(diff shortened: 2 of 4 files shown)


agilgur5:
Tests pass on all Node versions, and I temporarily added [a commit with passing tests](https://github.com/ezolenko/rollup-plugin-typescript2/pull/326/checks?check_run_id=6435461924) to check Node 17 specifically as well.

PR #329, opened 2022-05-14
test: 100% coverage in `get-options-overrides` (`createFilter`)

agilgur5:
## Summary

Get 100% coverage of `get-options-overrides` by adding tests for remaining parts of `createFilter`

Follow-up to https://github.com/ezolenko/rollup-plugin-typescript2/pull/321#discussion_r867520059 (plus more stuff too)

## Details

- get the skipped test to work by using absolute paths, as that is what the `filter` function expects (and is how it is used in this codebase)
  - use the same helper func on the main `createFilter` test as well to ensure that we're consistently testing the same paths
    - (though `**/*` basically matches _everything_)

- add a test for project references as well
  - and remove the `**/*` from the include for this so it doesn't match everything
    - this also tests the single string include code path as well

- add a test for the `context.debug()` statements as well
  - couldn't get to "100% Funcs" coverage without this
  - used a simplified mock instead of actually using `RollupContext` so that this doesn't rely on that file/unit to work

### DIFF
--- a/__tests__/get-options-overrides.spec.ts
+++ b/__tests__/get-options-overrides.spec.ts
@@ -1,4 +1,4 @@
-import { afterAll, test, expect } from "@jest/globals";
+import { afterAll, test, expect, jest } from "@jest/globals";
 import * as path from "path";
 import * as ts from "typescript";
 import { remove } from "fs-extra";
@@ -12,6 +12,9 @@ setTypescriptModule(ts);
 
 const local = (x: string) => path.resolve(__dirname, x);
 const cacheDir = local("__temp/get-options-overrides");
+
+// filter expects an absolute path and resolve include/exclude to process.cwd() by default: https://github.com/ezolenko/rollup-plugin-typescript2/pull/321#discussion_r873077874
+const filtPath = (path: string) => `${process.cwd()}/${path}`;
 
 afterAll(() => remove(cacheDir));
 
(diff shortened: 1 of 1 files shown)


__tests__/get-options-overrides.spec.ts
```diff
@@ -13,6 +14,9 @@ setTypescriptModule(ts);
 const local = (x: string) => path.resolve(__dirname, x);
 const cacheDir = local("__temp/get-options-overrides");
 
+// filter expects an absolute path and resolves include/exclude to process.cwd() by default: https://github.com/ezolenko/rollup-plugin-typescript2/pull/321#discussion_r873077874
+const filtPath = (relPath: string) => normalize(`${process.cwd()}/${relPath}`);
```
agilgur5, line 18: Tried both `path.normalize` (which will use the system's separators, same as `process.cwd()` I believe) and Rollup's `normalizePath` (which uses POSIX separators), and both failed...

`createFilter` uses `normalizePath` [under-the-hood](https://github.com/rollup/plugins/blob/36de7cc9c3915667bc0dd203ee7b139da8858959/packages/pluginutils/src/createFilter.ts#L16) as well.

Since only `rootDirs` and `projectReferences` failed, that got me thinking that maybe `expandIncludeWithDirs` is the issue, and indeed, it uses `path.join`, which uses the system's separators. So we get mixed separators again here, due to a bug in `expandIncludeWithDirs`: it needs to be normalized or use `posix.join` instead.

I'm gonna guess here that `rootDirs` and `projectReferences` aren't used very often (or at least not on Windows) by consumers, as this somehow hasn't come up as a bug report before

agilgur5:
@ezolenko this PR wasn't quite complete -- note that CI was failing on Windows.

I had made the in-line comment above to follow-up on, but hadn't had a chance to re-visit this. You marked the comment as "resolved", but it had not been resolved yet (hence why CI hadn't passed).

I intended on writing a Windows workaround for the test and then making a different, isolated PR to fix the source issue (as this one is intended to just be a testing PR) [here](https://github.com/ezolenko/rollup-plugin-typescript2/pull/329#discussion_r873108879) and for https://github.com/ezolenko/rollup-plugin-typescript2/pull/321#discussion_r869710856 (which are both in the same source file).

Since this was merged early, I'll try to get that PR fix out today so that CI passes on `master` on Windows.

I can't change PR settings myself, so you might want to change the settings to not allow PRs to be merged unless CI checks pass to prevent something like this in the future.

__tests__/get-options-overrides.spec.ts
agilgur5: Fixed this in #331

agilgur5:
@ezolenko I fixed the root cause of the Windows failures in #331

PR #331, opened 2022-05-26
fix: normalize paths in `get-options-overrides`

agilgur5:
## Summary

`get-options-overrides.ts` had some paths that were not normalized, resulting in mixed separators on Windows in certain cases.
- Fixes https://github.com/ezolenko/rollup-plugin-typescript2/pull/321#discussion_r869710856
- Fixes https://github.com/ezolenko/rollup-plugin-typescript2/pull/329#discussion_r873108879
  - Also follow-up to https://github.com/ezolenko/rollup-plugin-typescript2/pull/329#issuecomment-1138876847

## Details

- the `outDir` was not normalized after the `/placeholder` part was added to `cacheRoot`
  - `cacheRoot` could have `\` directory separators on it on Windows, which caused some tests to fail on Windows before
  - tests have been normalized now too

- `expandIncludeWithDirs` used `path.join` without normalizing after
  - `path.join` uses the OS's native separators (`posix.join` would do POSIX separators only), so when the paths were already normalized and then `path.join`ed, this would cause mixed separators on Windows
  - this fixes the current [CI failure on Windows](https://github.com/ezolenko/rollup-plugin-typescript2/runs/6516149780?check_suite_focus=true) in the `createFilter` tests (`rootDirs` and `projectReferences`, which use `expandIncludeWithDirs`)

## Review Notes

This should just be a bugfix and not breaking, but per the linked discussions, I'm really not sure how these bugs haven't been found yet by consumers -- I'm guessing that TS does normalization on `outDir` (lots of redundant normalization necessary due to ambiguous inputs) and that `rootDirs` and `projectReferences` just aren't used by many Windows users.

### DIFF
--- a/__tests__/get-options-overrides.spec.ts
+++ b/__tests__/get-options-overrides.spec.ts
@@ -11,7 +11,7 @@ import { getOptionsOverrides, createFilter } from "../src/get-options-overrides"
 
 setTypescriptModule(ts);
 
-const local = (x: string) => path.resolve(__dirname, x);
+const local = (x: string) => normalize(path.resolve(__dirname, x));
 const cacheDir = local("__temp/get-options-overrides");
 
 // filter expects an absolute path and resolves include/exclude to process.cwd() by default: https://github.com/ezolenko/rollup-plugin-typescript2/pull/321#discussion_r873077874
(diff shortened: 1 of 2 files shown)


agilgur5:
All tests pass on Windows now! Hoping we don't have to deal with path normalization again for a while now 😅

PR #333, opened 2022-05-30
- partially undoing lodash change

ezolenko:
Partially undoing recent lodash change (#328), `_.each()` semantics is apparently different from `Array.forEach()` -- during the build it created "undefined" folder with bunch of empty files.

### DIFF
--- a/src/tscache.ts
+++ b/src/tscache.ts
@@ -264,7 +264,7 @@ export class TsCache
 		if (this.ambientTypesDirty)
 			this.context.info(yellow("ambient types changed, redoing all semantic diagnostics"));
 
-		_.each(typeNames, (name) => this.typesCache.touch(name));
+		typeNames.forEach(this.typesCache.touch, this.typesCache);
 	}
 
 	private getDiagnostics(type: string, cache: ICache<IDiagnostics[]>, id: string, snapshot: tsTypes.IScriptSnapshot, check: () => tsTypes.Diagnostic[]): IDiagnostics[]
--- a/src/tscache.ts
+++ b/src/tscache.ts
@@ -180,13 +180,13 @@ export class TsCache
 
 		if (acyclic)
 		{
-			alg.topsort(this.dependencyTree).forEach(id => cb(id));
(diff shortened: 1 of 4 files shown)


src/tscache.ts
```diff
@@ -264,7 +264,7 @@ export class TsCache
 		if (this.ambientTypesDirty)
 			this.context.info(yellow("ambient types changed, redoing all semantic diagnostics"));
 
-		typeNames.forEach(this.typesCache.touch, this);
+		_.each(typeNames, (name) => this.typesCache.touch(name));
```
ezolenko, line 267: This line was responsible I think. I reverted all forEach changes just in case
image

agilgur5: the bug here might be that I didn't put the correct `this` context here when I re-wrote it -- let me double-check that shortly

agilgur5:
> during the build it created "undefined" folder with bunch of empty files.

ah my bad, thanks for catching this and the exact line too.

I did see this folder once or twice but thought that was because of all the other changes and investigations I was doing and just a temporary thing I broke -- turns out it wasn't temporary 😅 

> `_.each()` semantics is apparently different from `Array.forEach()`

The semantics are only _slightly_ different in that `_.each` handles falsey values, while `.forEach` doesn't -- but `.forEach` doesn't need to, as you can only use it on arrays anyway.

Another reason why I think it's a problem with `this` instead

src/tscache.ts
agilgur5: yep the problem is `this`. similar to the `Array#map` statements in `index.ts`, it should be:
```suggestion
		typeNames.forEach(this.typesCache.touch, this.typesCache);
```

Can confirm that works without issues

agilgur5 (changes requested):
See my suggested change to the line that causes this -- that's the only one that needs fixing and it's not a `forEach` issue, it's because of how I rewrote it (removing the anonymous arrow function). With that fix in, there should be no other issues with any lodash replacement.

So if we can just change _only_ that line, think this would be good to go!

Issue #97, opened 2018-07-07
ENOENT: no such file or directory, open '...?rollup-plugin-vue=script.d.ts'


gluons:
#### What happens and why it is wrong

Get this error when using with [`rollup-plugin-vue`](https://github.com/vuejs/rollup-plugin-vue) when set `declaration` to `true` in `tsconfig.json`.

```
Error: ENOENT: no such file or directory, open '<path to project directory>\dist\components\Hello.vue?rollup-plugin-vue=script.d.ts'
```

**Repo to Repro:** https://github.com/gluons/rollup-plugin-typescript2-issue-97

#### Environment 

**OS:** Windows 10
**Node:** v10.4.1
**Yarn:** 1.7.0

#### Versions
- typescript: 2.8.4
- rollup: 0.62.0
- rollup-plugin-typescript2: 0.15.1

#### rollup.config.js

<details>

```js
import { resolve } from 'path';

import ts from 'rollup-plugin-typescript2';
import vue from 'rollup-plugin-vue';

export default {
	input: resolve(__dirname, './src/index.ts'),
	output: [
		{
			file: resolve(__dirname, './dist/hello-plugin.umd.js'),
			format: 'umd',
			sourcemap: true,
			name: 'HelloPlugin',
			globals: {
				vue: 'Vue'
			}
		},
		{
			file: resolve(__dirname, './dist/hello-plugin.es.js'),
			format: 'es',
			sourcemap: true
		}
	],
	plugins: [
		ts({
			verbosity: 3 // Debug
		}),
		vue()
	],
	external: ['vue']
};
```

</details>

#### tsconfig.json

<details>

```json
{
	"compilerOptions": {
		"module": "es2015",
		"moduleResolution": "node",
		"target": "es5",
		"lib": [
			"dom",
			"es2015"
		],
		"esModuleInterop": true,
		"sourceMap": true,
		"declaration": true
	},
	"include": [
		"src"
	]
}
```

</details>

#### package.json

<details>

```json
{
  "name": "rolup-vue-ts-issue",
  "version": "0.0.0",
  "description": "A repo to repro Rollup + Vue + TypeScript with declaration bundling issue.",
  "private": true,
  "scripts": {
    "prebuild": "rimraf dist/*",
    "build": "rollup -c rollup.config.ts"
  },
  "author": "Saran Tanpituckpong <sarunta@gmail.com>",
  "license": "UNLICENSED",
  "devDependencies": {
[...]

Issue #129, opened 2019-01-09
Importing Vue SFC from `lang="ts"` block


eddow:
While switching from `rollup-plugin-typescript` to `rollup-plugin-typescript2` in order to produce declaration files, the `*.vue` files are not recognized anymore.

```
[!] (rpt2 plugin) Error: someFolder/index.ts(2,53): semantic error TS2307 Cannot find module './component.vue'.
src\index.ts
Error: someFolder/index.ts(2,53): semantic error TS2307 Cannot find module './component.vue'.
```

Trying by just importing `rollup-plugin-typescript` instead of `rollup-plugin-typescript2` bundles without a problem.

This could be bound to this [issue](https://github.com/ezolenko/rollup-plugin-typescript2/issues/78) though I have the last version (of every plugin today indeed).

ezolenko:
Could you post your tsconfig and rollup config?

ezolenko:
Or a small repo with reproduction :)

eddow:
I unfortunately don't have a "small" repo. I am working [here](https://github.com/eddow/vue-golden-layout/tree/rollup), trying to migrate from `webpack` to `rollup`.

The import in `rollup.config.js` can be changed to `rollup-plugin-typescript2` to see the difference.

danimoh:
Hello.

First of all thank you very much for working on this plugin. It does indeed make much more sense than `rollup-plugin-typescript`.

I can confirm that this issue exists and setup a small demo repository:
https://github.com/danimoh/rollup-plugin-typescript2-vue-demo

If you commment out the line `import AnotherComponent from './AnotherComponent.vue';` it does compile, but unfortunately not with this line enabled.

It's funny that we encountered this issue around the same time. Maybe it was caused by a recent change?
A guess from my side with very limited knowledge about rollup, rollup plugins and typescript would be:
Is it possible that typescript itself is trying to import `AnotherComponent` instead of rollup or `rollup-plugin-vue` handling that import first?

Any work on this issue is very much appreciated.

ezolenko:
Reproduced on both repos, not sure if it is the same problem yet, but it is likely.

To fix second case we do indeed need to send module resolution back to rollup so that vue plugin can do it's thing.

Problem with hooking up rollup's module resolution and typescript is that rollup in later versions returns a Promise from `context.resolveId(...)`. So the call chain looks like this:

- rollup calls plugin's transform
- plugin calls typescript's `LanguageService.getEmitOutput`
[...]

PR #336, opened 2022-05-31
fix: use `.d.ts` instead of `.vue.d.ts` for Vue declarations 

### LINKED ISSUE #224: Vue declaration files have wrong file endings -- `.vue.d.ts` instead of `.d.ts`
Building works fine but the generated declaration file from a SFC for example `MyComponent.vue` is named as `MyComponent.vue.d.ts` which breaks my imports (also generates like that without importing another SFC).

Renaming the file to `MyComponent.d.ts` works and the errors are gone.

#### Environment 
I'm using this plugin together with `rollup-plugin-vue` and `rollup-plugin-babel`. I also have `@rollup/plugin-commonjs` & `@rollup/plugin-node-resolve` if that does matter.
 
Could that be a configuration problem with my `tsconfig.json`? I will try to add the configs if needed.

#### Versions
- typescript: 3.8.3
- rollup: 2.3.2
- rollup-plugin-typescript2: 0.27.0

agilgur5:
## Summary

For the `rollup-plugin-vue` integration, output declarations as `name.d.ts` instead of `name.vue.d.ts`
- Fixes #224 , follow-up to https://github.com/ezolenko/rollup-plugin-typescript2/issues/224#issuecomment-642790321

## Details

- `.vue.d.ts` was recommended by the `rollup-plugin-vue` maintainers in https://github.com/ezolenko/rollup-plugin-typescript2/issues/97#issuecomment-405607066 back when this code was originally written, but apparently it causes problems with some imports according to #224
  - plain `.d.ts` seems to work fine according to users in that issue
    - and plain is the standard in TS too I believe?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -328,7 +328,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 
 				let fileName = entry.name;
 				if (fileName.includes("?")) // HACK for rollup-plugin-vue, it creates virtual modules in form 'file.vue?rollup-plugin-vue=script.ts'
-					fileName = fileName.split("vue?", 1) + extension;
+					fileName = fileName.split(".vue?", 1) + extension;
 
 				// If 'useTsconfigDeclarationDir' is given in the
 				// plugin options, directly write to the path provided
--- a/src/index.ts
+++ b/src/index.ts
@@ -328,7 +328,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 
 				let fileName = entry.name;
 				if (fileName.includes("?")) // HACK for rollup-plugin-vue, it creates virtual modules in form 'file.vue?rollup-plugin-vue=script.ts'
-					fileName = fileName.split("?", 1) + extension;
+					fileName = fileName.split("vue?", 1) + extension;
 
 				// If 'useTsconfigDeclarationDir' is given in the
 				// plugin options, directly write to the path provided


andrew0:
This change is breaking Vue type definitions for me. I put up a small repro here: https://github.com/andrew0/rollup-vue-ts-repro. The index.d.ts file contains

```ts
export { default as MyComponent } from './MyComponent.vue';
```

so the component's type definition should be `MyComponent.vue.d.ts`, not `MyComponent.d.ts`. I'm not really sure why the people in the original issue were having problems with the old behavior. Perhaps they were importing `.vue` files without an extension, but this was never recommended, and is [no longer supported by the Vue team](https://github.com/vitejs/vite/issues/178).

agilgur5:
> I'm not really sure why the people in the original issue were having problems with the old behavior.

@andrew0 this isn't the original issue, so I think it would be best to discuss that there.
I'm not a Vue user (and I don't believe ezolenko is either), so we've just been going with what users have asked us to implement. I count at least 4 people in that issue between comments and reactions (and most users do not interact with issues at all), so a non-trivial amount had asked for this.

> Perhaps they were importing `.vue` files without an extension, but this was never recommended, and is [no longer supported by the Vue team](https://github.com/vitejs/vite/issues/178).

It's possible, but there isn't a repro there to work off of and deduce. `rollup-plugin-vue` isn't supported either so this integration can be considered deprecated too. Not sure if Vue 2 / Vue 3 makes a difference either?

I see that issue got a decent amount of references (and probably has many more if it were not locked), so it sounds very possible that extensionless was pretty commonly used by a variety of users and was even supported by the CLI. It also mentions configuring Webpack for resolving the extension.
The original issue predates that Vite issue too
TS itself also commonly uses extensionless (Node ESM breaks that a bit though), so I can see this being pretty confusing to the community; whether or not that is the reasoning behind the original issue.

### next steps

I've dug into Vue a bit more in supporting rpt2; I think we'd probably want to follow whatever `vue-tsc` does ideally to follow "principle of least surprise" as we try to do with `tsc` already.
We'd also want to make sure the type definitions work with Volar -- which should hopefully function the same.

@andrew0 I think it would still be good for you to comment on the original issue to see what the users there think.

> I put up a small repro here: https://github.com/andrew0/rollup-vue-ts-repro.

There's another use-case that users might have been talking about too, importing Vue SFCs from each other. e.g. if you had `MyComponent2` imported by `MyComponent`, what would `vue-tsc` output and how would Volar interact with type definitions there?
Ideally the compiled files should work _without_ a Vue-specific tool like Volar etc, so we'd want to check compat on that too.
Lot of things to double-check if we want certainty 😕

agilgur5:
> I put up a small repro here: https://github.com/andrew0/rollup-vue-ts-repro.

I added `vue-tsc` and some additional config to your repro and confirmed that `vue-tsc` produces `.vue.d.ts` files, and _not_ `.d.ts` files.

I went ahead and created a Vue repro environment [here](https://stackblitz.com/edit/rpt2-vue-repro) for use for this and other Vue issues. That has updated config, uses Vue 3, etc, and similarly has `.vue.d.ts` from `vue-tsc`.

Based on this comparison, it seems like we may very well want to revert this PR in the next breaking release (`0.33.0`).

@andrew0 commenting on the original issue and getting clarification there would still be very helpful. Don't want to churn this more than it has already, and not sure if users there were using a common misconfiguration (well, now it is) or what

> Not sure if Vue 2 / Vue 3 makes a difference either?

It looked like you were [using Vue 2](https://github.com/andrew0/rollup-vue-ts-repro/blob/f1159e1e43740590b6153c6dea65d00f4db51da1/package.json#L16) in your repro, though I had the same results with Vue 3. Not sure if this might be because of how `vue-tsc` is set-up though.

I haven't yet checked in on Volar and other imports etc (mostly as I'm not a Vue user).

andrew0:
Thanks a lot for looking into this more, @agilgur5! I agree it makes sense to align with what `vue-tsc` does. I left a comment in the original issue to try to clarify why the filer was having issues with the old behavior.

>There's another use-case that users might have been talking about too, importing Vue SFCs from each other. e.g. if you had `MyComponent2` imported by `MyComponent`, what would `vue-tsc` output and how would Volar interact with type definitions there?

I modified your repro a bit to test how `vue-tsc` would handle this: https://stackblitz.com/edit/rpt2-vue-repro-2jy8kh. I think in most cases, importing `.vue` files from another `.vue` file works the same as importing a `.vue` file from a `.ts` file. One difference I did notice is that it seems that `vue-tsc` is able to merge the prop types when you extend components, whereas rpt2 does not.

`vue-tsc`:

```ts
declare const _default: import("vue").DefineComponent<{
    b: StringConstructor;
}, unknown, unknown, {}, {}, import("vue").ComponentOptionsMixin, import("vue").DefineComponent<{
    a: StringConstructor;
}, unknown, unknown, {}, {}, import("vue").ComponentOptionsMixin, import("vue").ComponentOptionsMixin, Record<string, any>, string, import("vue").VNodeProps & import("vue").AllowedComponentProps & import("vue").ComponentCustomProps, Readonly<import("vue").ExtractPropTypes<{
    a: StringConstructor;
}>>, {}>, Record<string, any>, string, import("vue").VNodeProps & import("vue").AllowedComponentProps & import("vue").ComponentCustomProps, Readonly<import("vue").ExtractPropTypes<{
    b: StringConstructor;
}>>, {}>;
export default _default;
```

`rollup-plugin-typescript2`:

```ts
declare const _default: import("vue").DefineComponent<{
    b: StringConstructor;
}, unknown, unknown, {}, {}, import("vue").ComponentOptionsMixin, any, Record<string, any>, string, import("vue").VNodeProps & import("vue").AllowedComponentProps & import("vue").ComponentCustomProps, Readonly<import("vue").ExtractPropTypes<{
    b: StringConstructor;
}>>, {}>;
export default _default;
```

So with `vue-tsc`, Volar would be able to give autocomplete suggestions for both the `a` and `b` props, whereas with rpt2, Volar would only see the `b` prop.

agilgur5:
> Thanks a lot for looking into this more, @agilgur5!

Thanks for engaging @andrew0 🙂 
We could definitely use help on the Vue front with this issue and a few others! If you have any time to look through the open Vue issues here (there's only 2 I believe), that could really help given our lack of Vue knowledge 🙁 

> So with `vue-tsc`, Volar would be able to give autocomplete suggestions for both the `a` and `b` props, whereas with rpt2, Volar would only see the `b` prop.

To clarify, Volar only picks this up with the `.vue.d.ts` extension, right?
As in, with rpt2 `0.32`'s `.d.ts` only, Volar doesn't work? But it works works with rpt2 `<0.32`?

Or does it work with both?

> One difference I did notice is that it seems that `vue-tsc` is able to merge the prop types when you extend components, whereas rpt2 does not.

Thanks for the example! I believe this illustrates an existing (and fairly complicated) issue, #129.
The difference here being that the `shims-vue.d.ts` file basically gives a "default" `.vue` type declaration, so TS doesn't completely fail to resolve the type as a result.

If I had to guess, `vue-tsc` is aware that files it processes are only TS or Vue, and therefore it can do Vue -> TS -> DTS or something (would need to look at the `vue-tsc` source more closely, given that the docs are fairly sparse).
Per #129, the difficulty with Rollup is that there can be any number of plugins on-chain, so rpt2 can't just delegate to a single one. That and that TS has a synchronous interface, whereas Rollup has an async one (would need to use something like [`deasync`](https://github.com/abbr/deasync) to support Rollup's `this.resolve` / `this.load` inside of the TS `LanguageServer`).
TS needs to have knowledge of the whole "Program"; it can't just compile per-file in order to do semantic type-checking or to produce declarations. So basically, when it sees a `.vue` file in the import chain, it needs to be able to resolve the types for that file as it may impact the file currently being processed. But rpt2 doesn't know what to do with `.vue` SFC files or how to otherwise get their types, causing #129. There's also some circular delegation required there, rpt2 -> rpv -> rpt2 etc.

agilgur5:
Reverted in #410 and released in [`0.33.0`](https://github.com/ezolenko/rollup-plugin-typescript2/releases/0.33.0.1)

PR #332, opened 2022-05-27
fix: add `realpath` to host to properly resolve monorepos

agilgur5:
## Summary

Implement `realpath` in `host.ts` to properly resolve symlinks, i.e. for monorepos

While the code for this is tiny, this is a pretty _huge_ bugfix given the number of users that have run into this here and downstream (in `microbundle`, TSDX, etc) -- glad to finally fix such a common error!

## Details

- tested this in a `pnpm` repo with symlinked deps and it worked there, so I believe this fixes all `pnpm` issues
  - it may also fix some Lerna issues if they were due to symlinks, but I didn't check those
  - not sure about others, e.g. Rush, Yarn workspaces, Yarn PnP

- I figured out this was needed by staring at the TS source code and then I found this [line](https://github.com/microsoft/TypeScript/blob/67673f324dd5f9398bb53fd16bf75efd155c32e7/src/compiler/moduleNameResolver.ts#L1412) 
  - it expects `host.realpath` to be implemented for TS's `realPath` to work correctly, otherwise it just returns the path with no transformation (i.e. the path to the symlink instead of the realpath)
    - this is not documented _anywhere_ and we were hitting this when calling `getEmitOutput`, before even using `moduleNameResolver`
  - so I just tried implementing it... and it worked!
  - notably, the other Rollup TS plugins don't implement this either???
    - not sure how they don't error on this??

- See my root cause analysis in https://github.com/ezolenko/rollup-plugin-typescript2/issues/234#issuecomment-1139202740 for (much) more details

## Review Notes

- note that I added a `!` as `realpath` doesn't have to be implemented on `ts.sys`... but it is in the default implementation (see comment)
  - I originally had a ternary with `fs.realpathSync` if it didn't exist but that is literally what the default implementation uses
    - can add this back in the future if it becomes an issue
    
## References

- Fixes #330
- Fixes #234
- ~Possible fix for #254, which mentions `pnpm` symlinks, but haven't tested~
  - **EDIT**: I investigated and responded in https://github.com/ezolenko/rollup-plugin-typescript2/issues/254#issuecomment-1140503934 -- that's actually completely unrelated and `pnpm` and other things were totally red herrings to the root cause
- Possible fix for #214 and #188, which are Lerna repos, but haven't tested
- Possible fix for #274, which mentions "monorepo" and has a similar kind of issue with declaration output as #330, but there is no repro to test against
- ~Also possible fix for #237, which has a workaround and mentions "monorepo", but unsure if related and no repro was provided there~
[...]

PR #340, opened 2022-06-03
!! [!important! merge and release me first!] deps: upgrade `@rollup/plugin-commonjs` to v22 to fix try/catch requires

### LINKED ISSUE #339: `ReferenceError: window is not defined` in 0.32.0
#### Environment

* Yarn 3

##### Versions

The command to use `envinfo` throws an error, so I'm resorting to copy-paste. All versions are latest.

```text
    "rollup": "^2.75.5",
    "rollup-plugin-typescript2": "^0.32.0",
    "typescript": "^4.7.2"
```

The error resides here:

```js
if (typeof commonjsRequire === "function") {
  try {
    lodash = {
      clone: require("lodash/clone"),
      constant: require("lodash/constant"),
      each: require("lodash/each"),
      filter: require("lodash/filter"),
      has:  require("lodash/has"),
      isArray: require("lodash/isArray"),
      isEmpty: require("lodash/isEmpty"),
      isFunction: require("lodash/isFunction"),
      isUndefined: require("lodash/isUndefined"),
      keys: require("lodash/keys"),
      map: require("lodash/map"),
      reduce: require("lodash/reduce"),
      size: require("lodash/size"),
      transform: require("lodash/transform"),
      union: require("lodash/union"),
      values: require("lodash/values")
    };
  } catch (e) {
    // continue regardless of error
  }
}

if (!lodash) {
  lodash = window._;
}
```

The line saying to use `window._` throws an error because `window` does not exist in the Node environment.

This is newly introduced in 0.32.0. Reverting to 0.31.0 fixes this.

agilgur5:
@ezolenko this is a bit important as 0.32.0 is currently broken for some users. Ironically, I suggested calling out dependency changes in https://github.com/ezolenko/rollup-plugin-typescript2/pull/332#issuecomment-1144411634 and this issue is in fact due to a breaking change in a dependency. In a dev dependency at that!

## Summary

Upgrade `@rollup/plugin-commonjs` to [v22.0.0](https://github.com/rollup/plugins/blob/master/packages/commonjs/CHANGELOG.md#v2200) to fix a breaking change in [v21.0.0](https://github.com/rollup/plugins/blob/master/packages/commonjs/CHANGELOG.md#v2100) that broke the bundling of `graphlib` and how it imports `lodash`, breaking rpt2 as a whole.
- Fixes #339 

## Details

- v21 broke [`graphlib`'s internal usage of `lodash`](https://github.com/ezolenko/rollup-plugin-typescript2/commit/08d2f5bcad7cb9a9a0974f3707a27d29dd89da7c#diff-80441564774ccf83c60a1616524830d5c15a9bc194b62e2c3d9f43de805c62d0L22435) with its breaking change, and v22 has another breaking change to fix that
  - notably, this was breaking rpt2 imports for some users because rpt2 bundles in `graphlib` etc, meaning the nuances of the bundling got leaked to consumers, and, in fact, broke their builds
    - another upside to unbundling, i.e. #80, is that this situation wouldn't happen
    
- run a build as well since that's necessary for/vital to this change

Can see more details in my root cause analysis in https://github.com/ezolenko/rollup-plugin-typescript2/issues/339#issuecomment-1145991074 and the comments leading up to that. https://github.com/rollup/plugins/pull/1005#issuecomment-954130548 is the central bug that applies to this case too.

## Review Notes

Note how _massive_ a change it is to the `dist` due to just changing the version of `@rollup/plugin-commonjs`. I can't even link to the line in the diff where `graphlib`'s usage of `lodash` changed because GitHub won't load a diff that big. I can confirm though that it does change that same area of code that was broken, and that testing this branch/PR in a user's repo resolved the issue.

I was able to repro in a user's library but couldn't repro this in rpt2 itself (when it bundles itself as in) or in one of my libraries though, so it's possibly an environment dependent bug or a race condition.

This will need a release once merged.


ezolenko:
Looks good (once you ignore all the whitespace :D). Do you think this should be 32.1 or 33.0?

CharlesStover:
>This will need a release once merged.

I'm not sure the release process for this repository. Should this commit include a patch version bump?

agilgur5:
> once you ignore all the whitespace :D

Ah that's a good tip, I did not realize how much was whitespace changes

> Do you think this should be 32.1 or 33.0? @agilgur5

@ezolenko `0.32.1` in my opinion, since `0.32.0` is partly broken due to this. It is a bugfix rather than a breaking change in my opinion. Also, just for reference, I try to point out things one may consider as breaking in PRs too.
This change should semantically make the dependencies work as they did in `0.31`, and I think that's the main breakage in `0.32.0`.
(The rest of the main changes are in `0.32.0` are "bugfixes that change some incorrect semantics", could consider all of them non-breaking as well).

> I'm not sure the release process for this repository. Should this commit include a patch version bump?

Generally there's a separate commit that ezolenko makes that _only_ bumps versions. That's done in most repos I've worked with, especially since a release may batch a few PRs together.

ezolenko:
Yep, release process goes roughly like this: 
- package version is set at least a patch version higher than previous release (usually this is done right after previous release)
- if the changes looks breaking, a minor version is incremented
- I do a build, review the diff and publish on npm
- build is then committed to git (if it wasn't already) and git release archive is made from that revision (so what's in npm and what's in git release under the same tag should match exactly)
- package version is then incremented

ezolenko:
That second point (about minor version bump) is a pure judgement call. Hopefully it would be more substantiated eventually if we get integrations tests and stuff. :)

ezolenko:
This is released in 0.32.1 btw

agilgur5:
> Yep, release process goes roughly like this:

It might be good to add your exact steps, specifically the commands you run, into a "Publishing" section (under "Development") in the new `CONTRIBUTING.md`.
So that way it's explicitly described and other people can replicate it (and in the future maybe, so that others can release too). Also from personal experience, I legit sometimes forget my own publishing process if I haven't released in a bit 😅 

> That second point (about minor version bump) is a pure judgement call. Hopefully it would be more substantiated eventually if we get integrations tests and stuff. :)

💯  integration tests will probably be the next big thing I tackle as they're _really_ starting to be necessary to make sure that some of the deeper bugfixes I'm making don't break a bunch of things.

But even then, version bumps are always subjective to an extent.
Even when I've tried to reason about a change with [ComVer](https://gitlab.com/staltz/comver) (which has no bugfixes), there are still some cases where something could be _considered_ breaking, but was intended as more of a feature.
Or, in a broader sense, what one identifies as "the API surface" is _itself_ subjective. For instance, a hypothetical: are thrown errors an API surface? In which case, changing an error message to be "more helpful" (i.e. a "feature") could also be considered breaking.

PR #311, opened 2022-04-21
fix(github): improve formatting of issue template spoilers

agilgur5:
## Summary

Improve formatting of the issue template with code blocks and fixing some duplication

## Details

- I originally added the spoilers in #244 and while they work (and I would say improve readability a good bit), there's also some issues with them
  - I've used them a ton more now so know how to workaround most of the issues with them now

- fix: use HTML `code` tag inside of `summary` tag, can't use backticks
  - was mixing MD and HTML before, and this doesn't always work and didn't work on GitHub, they just had visible backticks before

- fix: don't duplicate the file name in the heading, just make the `summary` have a heading inside it instead
  - use an `h4` same as the `####` that it was before

- feat: add syntax highlighting by adding code blocks for each code snippet
  - `js` for `rollup.config.js`, `json5` for `tsconfig` (it has comments, trailing commas, etc (actually a custom parser, but json5 is close enough)), `json` for `package.json`, and `text` for verbose logs
  - also, a lot of people sometimes just paste the code with no code block and it formats terribly, so this should help defer that (as well as any potential issues that can crop up with unindented blocks)
    - the `envinfo` `text` code block that I added in #245 seems to be working well and used in most issues, so hopefully this will improve issues too
    
## References

I wrote #310 using this format as an example! Think it's a good bit cleaner

- As for examples of how it previously looked and why this is better, #309 shows an example with no syntax highlighting but use of code blocks, whereas #306 uses no code blocks at all and so isn't formatted well and GitHub/Markdown doesn't indent it either. Both are properly using spoilers as well.

Issue #254, opened 2020-11-20
Vague `failed to transpile` error -- `noEmitOnError: true` breaks rpt2


benmccann:
#### What happens and why it is wrong

I get the error:

```
[!] (plugin rpt2) Error: failed to transpile 'packages/kit/src/runtime/navigation/index.ts'
```

This is almost impossible to debug. Luckily I figured out I can do `npx tsc` and get the real error message:
```
src/runtime/navigation/types.ts:33:9 - error TS2304: Cannot find name 'Route'.

33  route: Route;
           ~~~~~


Found 1 error.
```

##### Versions

```text
npx: installed 1 in 1.212s

  npmPackages:
    rollup: ^2.32.0 => 2.32.0 
    rollup-plugin-typescript2: ^0.29.0 => 0.29.0 
    typescript: ^4.0.3 => 4.0.3
```

#### Steps to reproduce

If you do not have `pnpm` then first run `npm install -g pnpm`

```sh
git clone git@github.com:sveltejs/kit.git
cd kit
git reset --hard 2483ae3730a7219814cc654c03b0409a797de2ee
pnpm install
pnpm -r build
```

After this, you can build just the relevant sub-project by running `cd packages/kit` and then `npx rollup -c` for faster debugging

#### Other context

Looks like it's coming from here: https://github.com/ezolenko/rollup-plugin-typescript2/blob/6fb0e75f5328666dca8ff7b11bc956096351a3eb/src/index.ts#L221

`printDiagnostics` right above looks to be being passed an empty array of diagnostics

agilgur5:
Per the downstream issue, it seems like OP believed this might be caused by #234 , with `pnpm` somehow swallowing the diagnostic errors

agilgur5:
I fixed `pnpm` (and symlinks in general) in #332, so I re-visited this to see if it solves the issue here, since `pnpm` was mentioned downstream.

It was a good thing that the [commit hash](https://github.com/sveltejs/kit/commits/2483ae3730a7219814cc654c03b0409a797de2ee) was included in the issue so that I could go back in time to the right spot.

I tried upgrading rpt2 to latest and added the fix in #332 locally to `node_modules`, but neither of those fixed it, errors were still not reported. I removed some of the rpt2 config to simplify things and added `clean: true`, but those changes didn't fix it either. I also set `verbosity: 3` (which the issue template requests the log from) and indeed no diagnostics are printed before this error.

[...]

Issue #211, opened 2020-03-03
Declarations not generated for type-only files not explicitly specified in `tsconfig`


slavafomin:
Hello, again!

I've noticed the following behavior: TypeScript declarations are not generated for source modules from which only types are imported and which are not explicitly added to the project using `files` or `include` options in `tsconfig.json`.

## Project configuration

```json5
// tsconfig.json
{
  "compilerOptions": {
    "outDir": "dist/",
    "declarationDir": "dist/types/",
    "declaration": true
  },
  
  // specifying only entry-point
  "files": ["src/index.ts"]
}
```

```javascript
// rollup.config.js

import typeScriptPlugin2 from 'rollup-plugin-typescript2';

export default {
  input: 'src/index.ts',
  output: {
    file: 'dist/index.js',
    format: 'esm',
  },
  plugins: [
    typeScriptPlugin2({
      verbosity: 2,
      clean: true,
      useTsconfigDeclarationDir: true,
    }),
  ],
};
```

## Source files

```typescript
// src/index.ts

// importing only type symbol from './foo.ts'
import { Foo } from './foo';

export class Main {
  constructor(foo: Foo) {
    console.log(foo);
  }
}
```

```typescript
// src/foo.ts

// type symbol
export type Foo = 'foo';

// constant symbol (not imported)
export const defaultFoo: Foo = 'foo';
```

```typescript
// src/unused.ts

export type Unused = 'unused';
```

## Generated files

```javascript
// dist/index.js

var Main = /** @class */ (function () {
    function Main(foo) {
        console.log(foo);
    }
    return Main;
}());

export { Main };
```

```typescript
// dist/types/index.d.ts

// Error: TS2307: Cannot find module './foo'.
import { Foo } from './foo';

export declare class Main {
    constructor(foo: Foo);
}
```

* the `dist/types/unused.d.ts` is not generated, which is **GOOD**.
* however, the `dist/types/foo.d.ts` is not generated, which is **BAD**.

---

And if I would set `"include": ["src/"]`, the `dist/types/unused.d.ts` would be generated, however, it shouldn't, because it's not imported anywhere.

---

Our goal here is to specify only the entry point of our project and let the compiler to handle only source files that are actually used in the project and avoid processing, generating declarations and reporting errors for files that are not used (redundant).

ezolenko:
That's supposed to work after #175, mostly anyway.

[...]

Issue #228, opened 2020-05-20
Cache is not being cleared when changing a transformer implementation


dominguezcelada:
#### What happens and why it is wrong
Without `clean: true` option, the plugin is catching the configuration even if changes at rollup config level are done (changing the implementation of a transformer implementation for instance)

Until verbosity: 3 was not set, it was not discovered cache could be an issue

##### Questions
* Is there any historical reason to set `clean: false` as default? It sounds like the default configuration should be without """magic""" on top right?
* What do you think about adding a warning when cache is used so users notice about it with default verbosity level?
* I did not check how cache is implemented but, is it an expected behaviour to not clean the cache when there are changes on the transformers passed to this plugin or is a bug?

#### Environment 
macOS catalina v10.15.4
node 14.2.0

#### Versions
- typescript: 3.9.3
- rollup: 2.10.5
- rollup-plugin-typescript2: 0.27.1

#### rollup.config.js
Not relevant

#### tsconfig.json
Not relevant

#### package.json
Not relevant

#### plugin output with verbosity 3
Not relevant

ezolenko:
This is a bug, the cache is supposed to catch rollup config changes -- the whole rollup config object is hashed and used as a part of cache key. Could you post your rollup config and what kind of change did you make? (or better yet, make a small repo with reproduction).

Thanks.

ezolenko:
Re warning about cache use, which output message made you realize cache might be involved? Maybe I need to print it on other verbosity levels...

dominguezcelada:
> This is a bug, the cache is supposed to catch rollup config changes -- the whole rollup config object is hashed and used as a part of cache key. Could you post your rollup config and what kind of change did you make? (or better yet, make a small repo with reproduction).
> 
> Thanks.

Sorry for the delay in the response. Here is a simplification of the problem:

```javascript
import { rollup } from 'rollup';
import typescript from 'rollup-plugin-typescript2';
import ts from 'typescript';

const nodeRemover = context => {
  return sourceFile => {
    const visitor = (node) => {
      if (ts.isImportDeclaration(node)) {
        return undefined;
      }

      return ts.visitEachChild(node, visitor, context);
    };

    return ts.visitNode(sourceFile, visitor);
  };
};

const transformer = () => ({
	before: [nodeRemover],
	after: []
});

rollup({
	input: './index.ts',
	plugins: [
		typescript({
			transformers: [transformer]
		})
	]
}).then(bundle => {
[...]

PR #345, opened 2022-06-05
fix: type-check `include`d files missed by `transform` (type-only files)

### LINKED ISSUE #298: type-checking doesn't work on type-only modules

When module contains only types (type,interface), and those types are being imported in other module, this plugin will omit those completely from type-checking which causes shipping invalid code to consumers.

🎬 Whole repro can be found here: https://stackblitz.com/edit/github-kgn16z-grsqrq?file=libs/test/rollup-config.js

1. run `yarn rollup:tsc` (uses this plugin with rollup) -> this will produce no Type errors 🚨 (Wrong)
2. run `yarn tsc:check`(uses raw tsc)  -> this will properly produce type errors ✅ (Correct, expected)

#### Source Code

```ts
// @filename src/index.ts
export * from './makeStyles';
```

```ts
// @filename src/makeStyles.ts

import { TestType } from './types';

export function makeStyles(): TestType {
  return {
    color: 'red',
  };
}
```

```ts
// @filename src/types.ts
interface CSS {
  backgroundColor: string;
}

export type TestType = {
  animationName?: CSS['background'];  // 🚨 this causes TS error
  color: string;
};
```
image

##### Versions

```text
typescript: ~4.4.3 => 4.4.4 
rollup@2.67.3
rollup-plugin-typescript2@0.31.2
```

#### rollup.config.js

<details>

```js
import * as path from 'path';
import ts from 'rollup-plugin-typescript2';
const repoRoot = path.resolve(__dirname, '..', '..');

const tsConfig = {
  check: true,
  tsconfig: 'libs/test/tsconfig.lib.json',
  verbosity: 4,
  tsconfigOverride: {
    compilerOptions: {
      rootDir: path.resolve(repoRoot, 'libs/test/src'),
      allowJs: false,
      declaration: true,
      declarationDir: path.resolve(repoRoot, 'out/tsc-dts'),
      paths: {
        test: ['libs/test/src/index.ts'],
      },
    },
  },
};

export default {
  input: path.resolve(__dirname, 'src/index.ts'),
  output: {
    file: path.resolve(repoRoot, 'out/bundle.js'),
    format: 'es',
  },
  plugins: [ts(tsConfig)],
};
```

</details>

#### tsconfig.json

<details>
  <summary>`tsconfig.json`: </summary>

```json
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "outDir": "../../dist/out-tsc",
    "types": ["node"]
  },
  "files": [
    "../../node_modules/@nrwl/react/typings/cssmodule.d.ts",
    "../../node_modules/@nrwl/react/typings/image.d.ts"
  ],
  "exclude": [
    "**/*.spec.ts",
    "**/*.test.ts",
    "**/*.spec.tsx",
    "**/*.test.tsx",
    "**/*.spec.js",
    "**/*.test.js",
    "**/*.spec.jsx",
    "**/*.test.jsx"
  ],
  "include": ["**/*.js", "**/*.jsx", "**/*.ts", "**/*.tsx"]
}

```

</details>

#### package.json

<details>
  <summary>`package.json`: </summary>

</details>

#### plugin output with verbosity 4

<details>
  <summary>plugin output with verbosity 3: </summary>

```
rpt2: built-in options overrides: {
    "noEmitHelpers": false,
    "importHelpers": true,
    "noResolve": false,
    "noEmit": false,
    "inlineSourceMap": false,
    "outDir": "/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2/placeholder",
    "moduleResolution": 2,
    "allowNonTsExtensions": true
}
rpt2: parsed tsconfig: {
    "options": {
        "rootDir": "/home/projects/github-kgn16z-grsqrq/libs/test/src",
        "sourceMap": true,
        "declaration": true,
        "moduleResolution": 2,
        "emitDecoratorMetadata": true,
        "experimentalDecorators": true,
        "importHelpers": true,
        "target": 2,
        "module": 99,
        "lib": [
            "lib.es2017.d.ts",
            "lib.dom.d.ts"
        ],
        "skipLibCheck": true,
        "skipDefaultLibCheck": true,
        "baseUrl": "/home/projects/github-kgn16z-grsqrq",
        "paths": {
            "test": [
                "libs/test/src/index.ts"
            ]
        },
        "pathsBasePath": "/home/projects/github-kgn16z-grsqrq/libs/test",
        "jsx": 4,
        "allowJs": false,
        "esModuleInterop": true,
        "forceConsistentCasingInFileNames": true,
        "strict": true,
        "noImplicitOverride": true,
        "noPropertyAccessFromIndexSignature": true,
        "noImplicitReturns": true,
        "noFallthroughCasesInSwitch": true,
        "outDir": "/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2/placeholder",
        "types": [
            "node"
        ],
        "declarationDir": "/home/projects/github-kgn16z-grsqrq/out/tsc-dts",
        "configFilePath": "/home/projects/github-kgn16z-grsqrq/libs/test/tsconfig.lib.json",
        "noEmitHelpers": false,
        "noResolve": false,
        "noEmit": false,
        "inlineSourceMap": false,
        "allowNonTsExtensions": true
    },
    "fileNames": [
        "/home/projects/github-kgn16z-grsqrq/node_modules/@nrwl/react/typings/cssmodule.d.ts",
        "/home/projects/github-kgn16z-grsqrq/node_modules/@nrwl/react/typings/image.d.ts",
        "/home/projects/github-kgn16z-grsqrq/libs/test/src/index.ts",
        "/home/projects/github-kgn16z-grsqrq/libs/test/src/makeStyles.ts",
        "/home/projects/github-kgn16z-grsqrq/libs/test/src/types.ts"
    ],
    "typeAcquisition": {
        "enable": false,
        "include": [],
        "exclude": []
    },
    "raw": {
        "extends": "./tsconfig.json",
        "compilerOptions": {
            "outDir": "../../dist/out-tsc",
            "types": [
                "node"
            ],
            "rootDir": "/home/projects/github-kgn16z-grsqrq/libs/test/src",
            "allowJs": false,
            "declaration": true,
            "declarationDir": "/home/projects/github-kgn16z-grsqrq/out/tsc-dts",
            "paths": {
                "test": [
                    "libs/test/src/index.ts"
                ]
            }
        },
        "files": [
            "../../node_modules/@nrwl/react/typings/cssmodule.d.ts",
            "../../node_modules/@nrwl/react/typings/image.d.ts"
        ],
        "exclude": [
            "**/*.spec.ts",
            "**/*.test.ts",
            "**/*.spec.tsx",
            "**/*.test.tsx",
            "**/*.spec.js",
            "**/*.test.js",
            "**/*.spec.jsx",
            "**/*.test.jsx"
        ],
        "include": [
            "**/*.js",
            "**/*.jsx",
            "**/*.ts",
            "**/*.tsx"
        ],
        "compileOnSave": false
    },
    "errors": [],
    "wildcardDirectories": {
        "/home/projects/github-kgn16z-grsqrq/libs/test": 1
    },
    "compileOnSave": false
}
rpt2: typescript version: 4.4.4
rpt2: tslib version: 2.3.1
rpt2: rollup version: 2.65.0
rpt2: rollup-plugin-typescript2 version: 0.31.1
rpt2: plugin options:
{
    "check": true,
    "tsconfig": "libs/test/tsconfig.lib.json",
    "useTsconfigDeclarationDir": true,
    "verbosity": 4,
    "tsconfigOverride": {
        "compilerOptions": {
            "rootDir": "/home/projects/github-kgn16z-grsqrq/libs/test/src",
            "allowJs": false,
            "declaration": true,
            "declarationDir": "/home/projects/github-kgn16z-grsqrq/out/tsc-dts",
            "paths": {
                "test": [
                    "libs/test/src/index.ts"
                ]
            }
        }
    },
    "clean": false,
    "cacheRoot": "/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2",
    "include": [
        "*.ts+(|x)",
        "**/*.ts+(|x)"
    ],
    "exclude": [
        "*.d.ts",
        "**/*.d.ts"
    ],
    "abortOnError": true,
    "rollupCommonJSResolveHack": false,
    "transformers": [],
    "tsconfigDefaults": {},
    "objectHashIgnoreUnknownHack": false,
    "cwd": "/home/projects/github-kgn16z-grsqrq",
    "typescript": "version 4.4.4"
}
rpt2: rollup config:
{
    "external": [],
    "input": "/home/projects/github-kgn16z-grsqrq/libs/test/src/index.ts",
    "plugins": [
        {
            "name": "rpt2"
        },
        {
            "name": "stdin"
        }
    ],
    "output": [
        {
            "file": "/home/projects/github-kgn16z-grsqrq/out/bundle.js",
            "format": "es",
            "plugins": []
        }
    ]
}
rpt2: tsconfig path: /home/projects/github-kgn16z-grsqrq/libs/test/tsconfig.lib.json
rpt2: included:
[
    "*.ts+(|x)",
    "**/*.ts+(|x)"
]
rpt2: excluded:
[
    "*.d.ts",
    "**/*.d.ts"
]
rpt2: Ambient types:
rpt2:     /home/projects/github-kgn16z-grsqrq/node_modules/@nrwl/react/typings/cssmodule.d.ts
rpt2:     /home/projects/github-kgn16z-grsqrq/node_modules/@nrwl/react/typings/image.d.ts
rpt2:     /home/projects/github-kgn16z-grsqrq/node_modules/@types/node/index.d.ts
rpt2: transpiling '/home/projects/github-kgn16z-grsqrq/libs/test/src/index.ts'
rpt2:     cache: '/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2/rpt2_4b2584b8fec2cb2070a57cc5fc1737a58e40ee79/code/cache/c054c5afccc541591135fb48ef93844cf144244d'
rpt2:     cache hit
rpt2:     cache: '/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2/rpt2_4b2584b8fec2cb2070a57cc5fc1737a58e40ee79/syntacticDiagnostics/cache/c054c5afccc541591135fb48ef93844cf144244d'
rpt2:     cache hit
rpt2:     cache: '/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2/rpt2_4b2584b8fec2cb2070a57cc5fc1737a58e40ee79/semanticDiagnostics/cache/c054c5afccc541591135fb48ef93844cf144244d'
rpt2:     cache hit
rpt2: generated declarations for '/home/projects/github-kgn16z-grsqrq/libs/test/src/index.ts'
rpt2: dependency '/home/projects/github-kgn16z-grsqrq/libs/test/src/makeStyles.ts'
rpt2:     imported by '/home/projects/github-kgn16z-grsqrq/libs/test/src/index.ts'
rpt2: resolving './makeStyles' imported by '/home/projects/github-kgn16z-grsqrq/libs/test/src/index.ts'
rpt2:     to '/home/projects/github-kgn16z-grsqrq/libs/test/src/makeStyles.ts'
rpt2: transpiling '/home/projects/github-kgn16z-grsqrq/libs/test/src/makeStyles.ts'
rpt2:     cache: '/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2/rpt2_4b2584b8fec2cb2070a57cc5fc1737a58e40ee79/code/cache/f1f8bd7cd1ae728a9fe42101454aeb9263d56bc2'
rpt2:     cache hit
rpt2:     cache: '/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2/rpt2_4b2584b8fec2cb2070a57cc5fc1737a58e40ee79/syntacticDiagnostics/cache/f1f8bd7cd1ae728a9fe42101454aeb9263d56bc2'
rpt2:     cache hit
rpt2:     cache: '/home/projects/github-kgn16z-grsqrq/node_modules/.cache/rollup-plugin-typescript2/rpt2_4b2584b8fec2cb2070a57cc5fc1737a58e40ee79/semanticDiagnostics/cache/f1f8bd7cd1ae728a9fe42101454aeb9263d56bc2'
rpt2:     cache hit
rpt2: generated declarations for '/home/projects/github-kgn16z-grsqrq/libs/test/src/makeStyles.ts'
rpt2: generating target 1
rpt2: rolling caches
rpt2: generating missed declarations for '/home/projects/github-kgn16z-grsqrq/node_modules/@nrwl/react/typings/cssmodule.d.ts'
rpt2: generating missed declarations for '/home/projects/github-kgn16z-grsqrq/node_modules/@nrwl/react/typings/image.d.ts'
rpt2: generating missed declarations for '/home/projects/github-kgn16z-grsqrq/libs/test/src/types.ts'
rpt2: emitting declarations for '/home/projects/github-kgn16z-grsqrq/libs/test/src/index.ts' to '/home/projects/github-kgn16z-grsqrq/out/tsc-dts/index.d.ts'
rpt2: emitting declarations for '/home/projects/github-kgn16z-grsqrq/libs/test/src/makeStyles.ts' to '/home/projects/github-kgn16z-grsqrq/out/tsc-dts/makeStyles.d.ts'
rpt2: emitting declarations for '/home/projects/github-kgn16z-grsqrq/libs/test/src/types.ts' to '/home/projects/github-kgn16z-grsqrq/out/tsc-dts/types.d.ts'
created out/bundle.js in 949ms
```

</details>

### Related issues

https://github.com/ezolenko/rollup-plugin-typescript2/issues/211

agilgur5:
## Summary

Type-check some type-only files that were missed in the `transform` hook but are in `parsedConfig.fileNames`
- Fixes #298 (which uses `include` globs)
- Fixes the latter half of #254 (which uses `include` globs)
- Partial workaround for #7, but not complete (see details below)
- Similarly partially fixes some things mentioned in #211, but does not _truly_/completely fix it as it uses `files` (see details below)
  - See my root cause analysis in https://github.com/ezolenko/rollup-plugin-typescript2/issues/298#issuecomment-1146658442 for how I plan to fix that
- Partially fixes the most common issue that I mentioned in https://github.com/ezolenko/rollup-plugin-typescript2/pull/311#issuecomment-1115491161, type-only imports

## Details

- type-only files never get processed by Rollup as their imports are elided/removed by TS in the resulting compiled JS file
  - so, as a workaround, make sure that all files in the `tsconfig` `include` (i.e. in `parsedConfig.fileNames`) are also type-checked
    - note that this will not catch _all_ type-only imports, in particular if one is using `tsconfig` `files` as in #211 (or in general _not_ using glob patterns in `include`) -- this is just a workaround, that requires a separate fix
  - we do the same process for generating declarations for "missed" files right now [in `_onwrite`](https://github.com/ezolenko/rollup-plugin-typescript2/blob/44116044750b01c3b0d612e36f453cc8e64f79b4/src/index.ts#L304), so basically do the same thing but for type-checking in ~`_ongenerate`~ **EDIT**: moved to [`buildEnd`](https://rollupjs.org/guide/en/#buildend) (see below)

(_Note_: _Technically_ speaking, there could be full TS files (i.e. _not_ type-only) that are in the `include` and weren't transformed. These would basically be independent TS files not part of the bundle that the user wanted type-checking and declarations for (as we _already_ generate declarations for those files))

For more details, see my root cause analysis in https://github.com/ezolenko/rollup-plugin-typescript2/issues/298#issuecomment-1146658442

## Review Notes

- ~This is built on top of #344, please merge that first. Marked this as a draft for now~
  - **EDIT**: rebased on top now that it was merged
- ~I'm using the `declarations` dict as a way to detect if a file has already run through the `transform` so as not to duplicate type-checking (both for performance and to not duplicate error messages)~. ~I'm not sure if this is optimal though, I merely copied `_onwrite`'s missed declaration check~.
  - **EDIT**: nvm, that wouldn't work given that you can type-check without outputting any declarations, created a new `Set` to track this instead  
  - ~This also runs for each Rollup `output`, so de-duping is pretty important in that sense too~
    - **EDIT**: moved to the `buildEnd` hook instead, so it doesn't run for each `output` anymore (and it doesn't generate any output files, unlike the missed declarations, so `buildEnd` is a more correct place for it as well)
   
    
- This exits successfully with zero on a type-error, while `tsc` will exit non-zero and not emit by default. `emitSkipped` is `false` too during the declaration output, despite the type-check error.
  - I could exit out with `this.error` in this case as well, but would have to be a different error message
 
## Misc

Could the `walkTree` call be moved into `buildEnd` as well? That seems like it might be a better place for it, the `!noErrors` output, and maybe `cache.done()` too, but I don't know if there was a reason for putting it in `_ongenerate`.

### DIFF
--- a/__tests__/integration/errors.spec.ts
+++ b/__tests__/integration/errors.spec.ts
@@ -1,4 +1,5 @@
 import { jest, afterAll, test, expect } from "@jest/globals";
+import { Mock } from "jest-mock"
 import * as path from "path";
 import { normalizePath as normalize } from "@rollup/pluginutils";
 import * as fs from "fs-extra";
@@ -12,7 +13,6 @@ jest.setTimeout(15000);
 
 const local = (x: string) => path.resolve(__dirname, x);
 const cacheRoot = local("__temp/errors/rpt2-cache"); // don't use the one in node_modules
-const onwarn = jest.fn();
 
 afterAll(async () => {
   // workaround: there seems to be some race condition causing fs.remove to fail, so give it a sec first (c.f. https://github.com/jprichardson/node-fs-extra/issues/532)
@@ -20,7 +20,7 @@ afterAll(async () => {
   await fs.remove(cacheRoot);
 });
 
-async function genBundle(relInput: string, extraOpts?: RPT2Options) {
+async function genBundle(relInput: string, extraOpts?: RPT2Options, onwarn?: Mock) {
   const input = normalize(local(`fixtures/errors/${relInput}`));
   return helpers.genBundle({
     input,
@@ -43,9 +43,10 @@ test("integration - semantic error", async () => {
 });
 
 test("integration - semantic error - abortOnError: false / check: false", async () => {
+  const onwarn = jest.fn();
   // either warning or not type-checking should result in the same bundle
-  const { output } = await genBundle("semantic.ts", { abortOnError: false });
-  const { output: output2 } = await genBundle("semantic.ts", { check: false });
+  const { output } = await genBundle("semantic.ts", { abortOnError: false }, onwarn);
+  const { output: output2 } = await genBundle("semantic.ts", { check: false }, onwarn);
   expect(output).toEqual(output2);
 
   expect(output[0].fileName).toEqual("index.js");
@@ -60,7 +61,38 @@ test("integration - syntax error", () => {
 });
 
 test("integration - syntax error - abortOnError: false / check: false", () => {
+  const onwarn = jest.fn();
   const err = "Unexpected token (Note that you need plugins to import files that are not JavaScript)";
-  expect(genBundle("syntax.ts", { abortOnError: false })).rejects.toThrow(err);
-  expect(genBundle("syntax.ts", { check: false })).rejects.toThrow(err);
+  expect(genBundle("syntax.ts", { abortOnError: false }, onwarn)).rejects.toThrow(err);
+  expect(genBundle("syntax.ts", { check: false }, onwarn)).rejects.toThrow(err);
+});
+
+const typeOnlyIncludes = ["**/import-type-error.ts", "**/type-only-import-with-error.ts"];
+
+test("integration - type-only import error", () => {
+  expect(genBundle("import-type-error.ts", {
+    include: typeOnlyIncludes,
+  })).rejects.toThrow("Property 'nonexistent' does not exist on type 'someObj'.");
+});
+
+test("integration - type-only import error - abortOnError: false / check: false", async () => {
+  const onwarn = jest.fn();
+  // either warning or not type-checking should result in the same bundle
+  const { output } = await genBundle("import-type-error.ts", {
+    include: typeOnlyIncludes,
+    abortOnError: false,
+  }, onwarn);
+  const { output: output2 } = await genBundle("import-type-error.ts", {
+    include: typeOnlyIncludes,
+    check: false,
+  }, onwarn);
+  expect(output).toEqual(output2);
+
+  expect(output[0].fileName).toEqual("index.js");
+  expect(output[1].fileName).toEqual("import-type-error.d.ts");
+  expect(output[2].fileName).toEqual("import-type-error.d.ts.map");
+  expect(output[3].fileName).toEqual("type-only-import-with-error.d.ts");
+  expect(output[4].fileName).toEqual("type-only-import-with-error.d.ts.map");
+  expect(output.length).toEqual(5); // no other files
+  expect(onwarn).toBeCalledTimes(1);
 });
--- a/__tests__/integration/fixtures/errors/import-type-error.ts
+++ b/__tests__/integration/fixtures/errors/import-type-error.ts
@@ -1,0 +1,8 @@
+// this file has no errors itself; it is used an entry file to test an error in a type-only import
+
+export type { typeError } from "./type-only-import-with-error";
+
+// some code so this file isn't empty
+export function sum(a: number, b: number) {
+  return a + b;
+}
--- a/__tests__/integration/fixtures/errors/type-only-import-with-error.ts
+++ b/__tests__/integration/fixtures/errors/type-only-import-with-error.ts
@@ -1,0 +1,2 @@
+type someObj = {};
+export type typeError = someObj['nonexistent'];
--- a/src/index.ts
+++ b/src/index.ts
@@ -287,6 +287,23 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 				});
 			}
 
+			// type-check missed files as well
+			parsedConfig.fileNames.forEach((name) =>
+			{
+				if (!pluginOptions.check)
+					return;
+
+				const key = normalize(name);
+				if (key in declarations || !filter(key)) // don't duplicate if it's already been transformed
+					return;
+
+				context.debug(() => `type-checking missed '${key}'`);
+
+				const snapshot = servicesHost.getScriptSnapshot(key);
+				if (snapshot)
+					typecheckFile(key, snapshot, context);
+			});
+
 			if (!watchMode && !noErrors)
 				context.info(yellow("there were errors or warnings."));
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -253,15 +253,32 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 
 		buildEnd(err)
 		{
-			if (!err)
+			if (err)
+			{
+				// workaround: err.stack contains err.message and Rollup prints both, causing duplication, so split out the stack itself if it exists (c.f. https://github.com/ezolenko/rollup-plugin-typescript2/issues/103#issuecomment-1172820658)
+				const stackOnly = err.stack?.split(err.message)[1];
+				if (stackOnly)
+					this.error({ ...err, message: err.message, stack: stackOnly });
+				else
+					this.error(err);
+			}
+
+			if (!pluginOptions.check)
 				return
 
-			// workaround: err.stack contains err.message and Rollup prints both, causing duplication, so split out the stack itself if it exists (c.f. https://github.com/ezolenko/rollup-plugin-typescript2/issues/103#issuecomment-1172820658)
-			const stackOnly = err.stack?.split(err.message)[1];
-			if (stackOnly)
-				this.error({ ...err, message: err.message, stack: stackOnly });
-			else
-				this.error(err);
+			// type-check missed files as well
+			parsedConfig.fileNames.forEach((name) =>
+			{
+				const key = normalize(name);
+				if (!filter(key))
+					return;
+
+				context.debug(() => `type-checking missed '${key}'`);
+
+				const snapshot = servicesHost.getScriptSnapshot(key);
+				if (snapshot)
+					typecheckFile(key, snapshot, context);
+			});
 		},
 
 		generateBundle(bundleOptions)
@@ -286,23 +303,6 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 						typecheckFile(id, snapshot, context);
 				});
 			}
-
-			// type-check missed files as well
-			parsedConfig.fileNames.forEach((name) =>
-			{
-				if (!pluginOptions.check)
-					return;
-
-				const key = normalize(name);
-				if (key in declarations || !filter(key)) // don't duplicate if it's already been transformed
-					return;
-
-				context.debug(() => `type-checking missed '${key}'`);
-
-				const snapshot = servicesHost.getScriptSnapshot(key);
-				if (snapshot)
-					typecheckFile(key, snapshot, context);
-			});
 
 			if (!watchMode && !noErrors)
 				context.info(yellow("there were errors or warnings."));
--- a/src/index.ts
+++ b/src/index.ts
@@ -292,6 +292,8 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 				});
 			}
 
+			const contextWrapper = new RollupContext(pluginOptions.verbosity, pluginOptions.abortOnError, this, "rpt2: ");
+
 			// type-check missed files as well
 			parsedConfig.fileNames.forEach((name) =>
 			{
@@ -303,7 +305,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 
 				const snapshot = servicesHost.getScriptSnapshot(key);
 				if (snapshot)
-					typecheckFile(key, snapshot, context);
+					typecheckFile(key, snapshot, contextWrapper);
 			});
 
 			buildDone();
--- a/src/index.ts
+++ b/src/index.ts
@@ -65,6 +65,15 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 				noErrors = false;
 	}
 
+	/** to be called at the end of Rollup's build phase, before output generation */
+	const buildDone = (): void =>
+	{
+		if (!watchMode && !noErrors)
+			context.info(yellow("there were errors or warnings."));
+
+		cache().done();
+	}
+
 	const pluginOptions: IOptions = Object.assign({},
 		{
 			check: true,
@@ -204,9 +213,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 				{
 					// always checking on fatal errors, even if options.check is set to false
 					typecheckFile(id, snapshot, contextWrapper);
-
 					// since no output was generated, aborting compilation
-					cache().done();
 					this.error(red(`failed to transpile '${id}'`));
 				}
 
@@ -259,6 +266,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 		{
 			if (err)
 			{
+				buildDone();
 				// workaround: err.stack contains err.message and Rollup prints both, causing duplication, so split out the stack itself if it exists (c.f. https://github.com/ezolenko/rollup-plugin-typescript2/issues/103#issuecomment-1172820658)
 				const stackOnly = err.stack?.split(err.message)[1];
 				if (stackOnly)
@@ -268,7 +276,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 			}
 
 			if (!pluginOptions.check)
-				return
+				return buildDone();
 
 			// walkTree once on each cycle when in watch mode
 			if (watchMode)
@@ -298,10 +306,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 					typecheckFile(key, snapshot, context);
 			});
 
-			if (!watchMode && !noErrors)
-				context.info(yellow("there were errors or warnings."));
-
-			cache().done();
+			buildDone();
 		},
 
 		generateBundle(this, _output)
--- a/src/index.ts
+++ b/src/index.ts
@@ -56,13 +56,13 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 
 	const typecheckFile = (id: string, snapshot: tsTypes.IScriptSnapshot, tcContext: IContext) =>
 	{
+			checkedFiles.add(id); // must come before print, as that could bail
+
 			const diagnostics = getDiagnostics(id, snapshot);
 			printDiagnostics(tcContext, diagnostics, parsedConfig.options.pretty !== false);
 
 			if (diagnostics.length > 0)
 				noErrors = false;
-
-			checkedFiles.add(id);
 	}
 
 	const pluginOptions: IOptions = Object.assign({},
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 import { relative, dirname, normalize as pathNormalize, resolve } from "path";
 import * as tsTypes from "typescript";
-import { PluginImpl, PluginContext, InputOptions, OutputOptions, TransformResult, SourceMap, Plugin } from "rollup";
+import { PluginImpl, InputOptions, TransformResult, SourceMap, Plugin } from "rollup";
 import { normalizePath as normalize } from "@rollup/pluginutils";
 import { blue, red, yellow, green } from "colors/safe";
 import findCacheDir from "find-cache-dir";
@@ -89,7 +89,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 	}
 	setTypescriptModule(pluginOptions.typescript);
 
-	const self: Plugin & { _ongenerate: () => void, _onwrite: (this: PluginContext, _output: OutputOptions) => void } = {
+	const self: Plugin = {
 
 		name: "rpt2",
 
@@ -270,6 +270,20 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 			if (!pluginOptions.check)
 				return
 
+			// walkTree once on each cycle when in watch mode
+			if (watchMode)
+			{
+				cache().walkTree((id) =>
+				{
+					if (!filter(id))
+						return;
+
+					const snapshot = servicesHost.getScriptSnapshot(id);
+					if (snapshot)
+						typecheckFile(id, snapshot, context);
+				});
+			}
+
 			// type-check missed files as well
 			parsedConfig.fileNames.forEach((name) =>
 			{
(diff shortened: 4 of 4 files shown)


agilgur5:
@ezolenko this is ready for review now. I double-checked that this worked locally against a similar error to #298's example (still need to get to adding integration tests).
If you could respond to the last review note and the "Misc" questions as well, that would be helpful

ezolenko:
I think the main reason walkTree was in `_ongenerate` was that there was no `buildEnd`  hook at the moment (or I missed it). That's why there is `generateRound` variable too, to avoid reprinting errors it for each output target.

If it works in `buildEnd` it can go there. Main requirement for watch mode is that all errors are printed at the end of each cycle, even for files that were not retranspiled in a given cycle. 

If you run in watch mode and break / fix / break / fix various files you would see if this behaves as expected. This applies to return code as well, ideally error caught at that stage behaves like error caught at transpile stage regarding rollup behavior and return code. For example with `abortOnError` option set, it should print first error and bail out without typechecking anything else (I imagine this option being used when failing must happen as soon as possible).

agilgur5:
> I think the main reason walkTree was in `_ongenerate` was that there was no `buildEnd` hook at the moment (or I missed it). That's why there is `generateRound` variable too, to avoid reprinting errors it for each output target.

OOOHHHH... that makes so much more sense. It looked like a workaround for `buildEnd`, but for some reason I didn't suspect that `buildEnd` literally _didn't exist_ before. 

Confirmed in the Rollup changelog that `buildEnd` was added in [`0.60`](https://github.com/rollup/rollup/blob/master/CHANGELOG.md#0600), which this code indeed predates.
rpt2 currently sets a min version of Rollup [`>=1.26.3`](https://github.com/ezolenko/rollup-plugin-typescript2/blob/63a644a7620941936378d112cdffbce529bcfb50/package.json#L41), so it wouldn't require any version changes to switch to using `buildEnd` instead now.

I think some of the other legacy checks like [`_.isFunction(this.error)`](https://github.com/ezolenko/rollup-plugin-typescript2/blob/63a644a7620941936378d112cdffbce529bcfb50/src/index.ts#L208) and [`if (this.addWatchFile)`](https://github.com/ezolenko/rollup-plugin-typescript2/blob/63a644a7620941936378d112cdffbce529bcfb50/src/index.ts#L222) can be removed now too. I thought they were still necessary for backward-compat, but all those seem to predate even Rollup 1.0.0.


> If you run in watch mode and break / fix / break / fix various files you would see if this behaves as expected.

Let me double-check this, but it should behave like that according to the Rollup spec.

> Main requirement for watch mode is that all errors are printed at the end of each cycle, even for files that were not retranspiled in a given cycle.

Oh, is this meant as a workaround for watch mode clearing the terminal by default on a new cycle?
I thought the optimal tree walk would only be for files that depend on those that changed. Since `transform` isn't called on every file again, at least, as far as I understand (and I may have an incorrect understanding of that), so lots of errors still won't be reported or declarations re-generated as a result of that.

> This applies to return code as well, ideally error caught at that stage behaves like error caught at transpile stage regarding rollup behavior and return code. For example with `abortOnError` option set, it should print first error and bail out without typechecking anything else (I imagine this option being used when failing must happen as soon as possible).

So I didn't go into this into _that_ much detail in "Review Notes", but the type-errors here seem to never return `emitSkipped` (probably because they don't impact the `Program` except for ambient typings? not sure). And in `transform` we only bail out when `emitSkipped: true`.
So the current `buildEnd` code _does_ act in the same way as `transform`, but `tsc` doesn't act this way -- which is _very_ confusing -- `tsc` _does_ exit non-zero, despite `emitSkipped: false`. Like it outputs declarations, and those declarations don't pass type-check (same as the type-only file itself), so `emitSkipped: false`, but `tsc` exits non-zero.
It only doesn't emit when `noEmitOnError: true` (the opposite of the default).

So I could change `buildEnd` to call `this.error`, but that behavior is actually slightly different from `transform`.
Like a key difference would be for `check: false`: there's no way for us to know to bail and attempt type-checking on a fatal error because `emitSkipped: false`. I.e. the flag for a fatal error _itself_ isn't set by TS.
So if we do call `this.error` in `buildEnd`, this would mean that it may bail if `check: true`, but will pass if `check: false`, even with `abortOnError: true` in both cases.

That's some weird divergent behavior, so I'm not sure which is optimal. But I can do either, just want to call out those trade-offs explicitly because this is unique to these files.

> For example with `abortOnError` option set, it should print first error and bail out without typechecking anything else (I imagine this option being used when failing must happen as soon as possible).

Kinda unrelated, but this has popped up in the issues pretty often since TS defaults to [`noEmitOnError: false`](https://www.typescriptlang.org/tsconfig#noEmitOnError).
Implementing #168 would be ideal so that we can more closely match `tsc`'s behavior, but I'm not sure if that'll always produce valid JS code, since Rollup could throw, which kinda ruins the point of improving DX, as it would be worse DX for Rollup to throw a confusing error. I looked into your old branch for that one and I think it doesn't produce invalid code, and if `emitSkipped`, we could just `return null` or something in `transform`, then bail out in `buildEnd` instead, once _all_ errors have printed out. So I think it's doable, but not 100% sure on that.

agilgur5:
> Let me double-check this, but it should behave like that according to the Rollup spec.

Yep, `buildEnd` behaves as expected.

The only problem I discovered during watch mode is that some errors can get printed twice 😕  . That's because the new `buildEnd` type-checks and `walkTree` type-checks as well. And `walkTree` intentionally doesn't check `checkedFiles`.

So I'm gonna have to switch the ordering here. Which means I might as well move `_ongenerate` into `buildEnd` in this PR

agilgur5:
Ah, `buildEnd` also has one _tiny_ difference from the current methodology -- `buildEnd` is called even if a build errors out, whereas `generateBundle` / `_ongenerate` isn't.

So I need to add a check for the error case and not `walkTree` if so. But still type-check `parsedConfig.fileNames` to support #168 (which I think should be the default to match `tsc`).

agilgur5:
> And in `transform` we only bail out when `emitSkipped: true`.

Ok, turns out this isn't entirely correct. I thought my understanding of this was a bit off.

Since `abortOnError` is [baked into](https://github.com/ezolenko/rollup-plugin-typescript2/blob/63a644a7620941936378d112cdffbce529bcfb50/src/rollupcontext.ts#L37) the `RollupContext`, a TS diagnostic of [`DiagnosticCategory.Error`](https://github.com/ezolenko/rollup-plugin-typescript2/blob/63a644a7620941936378d112cdffbce529bcfb50/src/print-diagnostics.ts#L21) will cause it to bail.

But in this portion of code we're using `ConsoleContext` instead of `RollupContext`, so I think therein lies the problem.

I actually have been wondering about why `ConsoleContext` vs. `RollupContext` (which is _only_ used in `transform`) for a while -- could you clarify that @ezolenko ?
one of the few missing pieces in my current understanding of the codebase

agilgur5:
> The only problem I discovered during watch mode is that some errors can get printed twice 😕 . That's because the new `buildEnd` type-checks and `walkTree` type-checks as well. And `walkTree` intentionally doesn't check `checkedFiles`.
> 
> So I'm gonna have to switch the ordering here. Which means I might as well move `_ongenerate` into `buildEnd` in this PR

Completed moving `_ongenerate` into `buildEnd` now and fixing the ordering. While I was at it, basically renamed `_onwrite` to `generateBundle`, as now that's the only function called in `generateBundle` anyway.

> Ah, `buildEnd` also has one _tiny_ difference from the current methodology -- `buildEnd` is called even if a build errors out, whereas `generateBundle` / `_ongenerate` isn't.

Handled the error case as well now.

> But in this portion of code (both the new code I added and the existing `walkTree` code) we're using `ConsoleContext` instead of `RollupContext`, so I think therein lies the problem.
> 
> I actually have been wondering about why `ConsoleContext` vs. `RollupContext` (which is _only_ used in `transform`) for a while -- could you clarify that @ezolenko? Should `RollupContext` be used instead of `ConsoleContext` in this new code and in the existing `walkTree` code?

So this is the only portion left remaining for this PR, that I'm actually not sure of myself (as it is historical, as far as I know), so could still use some input on this to finish up this PR

ezolenko:
The only reason `ConsoleContext` exists is that `PluginContext` is not available in some places where we want to print something. And the main behavior difference is that printing error in `PluginContext` would kill the build , but printing error in `ConsoleContext` would not, that's why `RollupContex` has this `bail` check to downgrade error to warning. There is also some cosmetic output differences I think.

agilgur5:
> The only reason `ConsoleContext` exists is that `PluginContext` is not available in some places where we want to print something.

Yea, that's what I thought, but looking at it _currently_, I'm pretty sure `PluginContext` exists in every usage of `ConsoleContext`. So maybe that's a historical remnant too, before Rollup made it exist in every hook.

`RollupContext` also has a [`hasContext`](https://github.com/ezolenko/rollup-plugin-typescript2/blob/8334c9b50fec01d9094b90b71a0913d4b87f94e1/src/rollupcontext.ts#L35) var that it checks against as well, and if falsey, it falls back to the same behavior as `ConsoleContext` anyway.
And I remove that var in #374, because that's never the case anymore. So definitely seems like a historical remnant.

So I believe `ConsoleContext` is _entirely_ redundant nowadays.

> And the main behavior difference is that printing error in `PluginContext` would kill the build , but printing error in `ConsoleContext` would not, that's why `RollupContext` has this `bail` check to downgrade error to warning

Yea that was a central difference, but at least in `index`, `ConsoleContext` is never even used when an error might be displayed.
It is passed into some other functions, namely `printDiagnostics` for `tsconfig` checks, that _could_ result in an error though. But in those cases, I feel like we probably should bail out anyway -- a `tsconfig` error or `compilerOptions` error breaks a bunch of stuff, so erroring out makes sense to me.
Let me know if you think otherwise though.

> There is also some cosmetic output differences I think.

Yea with `warn` and `error`, Rollup adds some formatting + coloring and plugin name + `[!]` or `(!)` as a prefix. For warnings, the [`onwarn`](https://rollupjs.org/guide/en/#onwarn) callback is also triggered.

But, per above, I don't think those differences are ever meaningful either.

So I'm pretty confident now that `ConsoleContext` is redundant / no longer necessary.

But if agreed, I'll remove it in a separate PR. For here, I'll just use `RollupContext` instead

agilgur5:
> Yea, that's what I thought, but looking at it _currently_, I'm pretty sure `PluginContext` exists in every usage of `ConsoleContext`. So maybe that's a historical remnant too, before Rollup made it exist in every hook.

> It is passed into some other functions, namely `printDiagnostics` for `tsconfig` checks, that _could_ result in an error though. But in those cases, I feel like we probably should bail out anyway -- a `tsconfig` error or `compilerOptions` error breaks a bunch of stuff, so erroring out makes sense to me.
> Let me know if you think otherwise though.

Ah, nevermind, it looks like [`options`](https://rollupjs.org/guide/en/#options) is indeed _still_ the exception to the rule.

So that means that maybe I shouldn't change `hasContext` in #374, and instead just remove `ConsoleContext`, since with the `hasContext` check, they have identical behavior when no `PluginContext`

ezolenko:
I think some other PRs I merged today broke this one

agilgur5:
Yea #373 was expected to conflict with this one (as I noted there) due to both using `buildEnd`, so one or the other was going to need to be updated.

I also still need to make that `RollupContext` change, then this should be good to go

agilgur5:
@ezolenko I think this PR is ready to go now. I fixed the merge conflict, used `RollupContext` (only for this new check, I didn't change `walkTree` for now), and also added an integration test for this.

CI is currently failing due to #377, not the code from this PR.

Also happened to stumble upon yet another bug (or two?) in an upstream codebase with this test: https://github.com/facebook/jest/issues/7136 roughly describes it, but basically mocks have poor isolation. Was able to workaround that one with a small refactor. Couldn't workaround another one, that mocks seem to lose state if an error is hit; see my commit message for more details.

Unrelated, but @ezolenko can you give a 👍  / 👎  on the proposal in #228? Have a contributor waiting for approval from you before tackling. (for reference, the proposal is mine, the contributor was going to implement it).

agilgur5:
> Ah, nevermind, it looks like [`options`](https://rollupjs.org/guide/en/#options) is indeed _still_ the exception to the rule.

`options` is still the exception to the rule, but we don't modify the `options` at all, so we can actually use [`buildStart`](https://rollupjs.org/guide/en/#buildstart), which was introduced at the same time as `buildEnd` in Rollup [`v0.60`](https://github.com/rollup/rollup/blob/master/CHANGELOG.md#0600).

Somehow didn't think of that till now, but that should obviate the need for `ConsoleContext` and simplify error code so that it all uses `context.error` instead of some code `throw`ing instead when that wasn't available.

PR #355, opened 2022-06-11
refactor(cache): `makeName` -> `createHash` for clarity

agilgur5:
## Summary

Rename `makeName` -> `createHash` and `name` -> `hash` in `tscache.ts` code for clarity

## Details

- I've actually been confused multiple times as to what `makeName` does when I read through the cache
  - I re-read the code and then am like "oh it's the hash"...
  - so thought renaming it to `createHash` would make things **a lot** clearer
    - `Name` -> `Hash`
    - `make` -> `create` because that's the more common terminology in programming

- also rename variables that reference `makeName`'s return from `name` to `hash`
  - for the same reason around clarity -- this way it's quicker to interpret whenever you see it too
  - not to mention, `name` can be confusing since we also have `id`, which is a path that is very similar to a name too
    - and lots of `fileName`s too
    - so good to disambiguate/differentiate a bit
  
## Misc Notes

Been digging into cache bugs more now that I've fixed a ton of the non-cache ones, so doing a bit of refactoring here as I'm going through it more frequently and in more depth.

### DIFF
--- a/src/tscache.ts
+++ b/src/tscache.ts
@@ -208,18 +208,18 @@ export class TsCache
 			return transform();
 		}
 
-		const name = this.makeName(id, snapshot);
+		const hash = this.createHash(id, snapshot);
 
 		this.context.info(`${blue("transpiling")} '${id}'`);
-		this.context.debug(`    cache: '${this.codeCache.path(name)}'`);
+		this.context.debug(`    cache: '${this.codeCache.path(hash)}'`);
 
-		if (this.codeCache.exists(name) && !this.isDirty(id, false))
+		if (this.codeCache.exists(hash) && !this.isDirty(id, false))
 		{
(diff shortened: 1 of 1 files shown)


ezolenko:
Really it's more like a key, with additional requirement of being filesystem safe, but this works too

agilgur5:
> Really it's more like a key

Hashes _are_ keys, so I think that terminology works 👍 



> with additional requirement of being filesystem safe

Ah, that's a good point that I didn't think about, but I guess a `sha1` hash is FS-safe, so it works so long as the implementation isn't changed.
Might be good to add a typedoc `@returns` comment mentioning that for future readers or anyone who may want to change the hash algorithm (I don't have any plans to do that, but I can see someone look at `sha1` and say "let's update that" and not be aware that FS-safe is a required property)

PR #363, opened 2022-06-22
refactor: use optional chaining for `sourceMapCallback`

agilgur5:
## Summary

Tiny one-liner change to use optional chaining
- following other optional chaining simplifications like #356, #335

## Misc

Is `sourceMapCallback` actually used by anyone? It's not listed in the docs so I didn't know it was a thing until I saw it in the code

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -251,8 +251,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 
 			if (result.map)
 			{
-				if (pluginOptions.sourceMapCallback)
-					pluginOptions.sourceMapCallback(id, result.map);
+				pluginOptions.sourceMapCallback?.(id, result.map);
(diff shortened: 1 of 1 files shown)


agilgur5:
> Is `sourceMapCallback` actually used by anyone? It's not listed in the docs so I didn't know it was a thing until I saw it in the code

Seems like it originates from #69 and that is the only reference to it in the entire issue/PR landscape, so that contributor may be the only person to use it 😅 

The use-case may also no longer be necessary with newer Istanbul versions (and `nyc` existing) based on some of the upstream comments linked, or with better source map handling in Rollup, but really hard to tell with such a complicated use-case without a simple repro

ezolenko:
The problem with having many options is that once they are in, they tend to stay in, unless you have lots of telemetry and actively trim unused ones...

PR #369, opened 2022-06-26
fix(cache): invalidate `codeCache` in most cases when imports change

### LINKED ISSUE #292: Cache not invalidating declarations when import changes (need to use `clean: true`)

I have a React project with compound componetns. Meaning - I'm using this kind of declaration:

```javascript
import {Child1} from './child1';
import {Child2} from './child2';

export const Parent = ({children}:{children:ReactNode}): JSX.Element => {
  return (
    <div>
      {children}
   </div>
  );
}

Parent.Child1 = Child1
Parent.Child2 = Child2
```

An example of usage:
```javascript
<Parent>
  <Parent.Child1 prop1="val1"/>
</Parent>
```

Now, if I change Child1's signature, for example changing it to expect `prop2` instead of `prop1` and then rebuilding, the `Child1.d.ts` file is updated but the `Parent.d.ts` file isn't. This means that I get exceptions when running the code because it still accepts the old `prop1`.

**This means I need to run all builds with `clean: true`**

##### Versions

```text
rollup: 2.60.2
rollup-plugin-typescript2: 0.30.0
typescript: 4.5.2
```

#### rollup.config.js

<details>

```js
// noinspection JSUnusedGlobalSymbols

import commonjs from "@rollup/plugin-commonjs";
import resolve from "@rollup/plugin-node-resolve";
import external from "rollup-plugin-peer-deps-external";
import image from '@rollup/plugin-image';
import typescript from "rollup-plugin-typescript2";

// @ts-ignore
import pkg from "./package.json";

export default {
    input: "./src/index.ts",
    output: [
        {
            file: pkg.main,
            format: "cjs",
            sourcemap: true
        },
        {
            file: pkg.module,
            format: "esm",
            sourcemap: true
        }
    ],
    plugins: [
        typescript({
            clean: true
        }),
        image(),
        external({
            includeDependencies: true
        }),
        resolve(),
        commonjs()
    ]
};
```

</details>

#### tsconfig.json

<details>
  <summary> `tsconfig.json` : </summary>

```json5
{
  "compilerOptions": {
    "baseUrl": "src",
    "rootDir": "src",
    "composite": true,
    "declaration": true,
    "declarationMap": true,
    "outDir": "dist",
    "declarationDir": "build",
    "module": "esnext",
    "target": "es5",
    "lib": [
      "es6",
      "dom",
      "es2016",
      "es2017"
    ],
    "sourceMap": true,
    "jsx": "react-jsx",
    "resolveJsonModule": true,
    "moduleResolution": "node",
    "esModuleInterop": true,
    "typeRoots": [
      "src/@types",
      "node_modules/@types"
    ]
  }
  },
  "include": [
    "src/**/*"
  ],
  "exclude": [
    "node_modules",
    "build"
  ]
}
```

</details>

agilgur5:
## Summary

The `codeCache` should be invalidated when imports change if declarations are needed, because imports' types can change the resulting declaration. 
- Fixes #292

**EDIT**: also added `!isolatedModules` to the check, see below comments regarding, e.g. enums, that could cause compiled JS code to change as well when imports change

## Details

- previously, the `checkImports` flag was set to `true` for type-checking, but `false` for compilation
  - I _believe_ it is `false` because the compiled JS shouldn't change if an import changes
    - though I'm not sure if that was the original intent behind the code
  - problematically though, compilation results can include declarations, and those _can_ change if imports change
    - for instance, the types of an import can change the declaration that is output
  - so now, only set it to `false` for compilation if declarations are _not_ needed

See my root cause analysis in https://github.com/ezolenko/rollup-plugin-typescript2/issues/292#issuecomment-1152951087

### DIFF
--- a/src/tscache.ts
+++ b/src/tscache.ts
@@ -213,8 +213,8 @@ export class TsCache
 	public getCompiled(id: string, snapshot: tsTypes.IScriptSnapshot, transform: () => ICode | undefined): ICode | undefined
 	{
 		this.context.info(`${blue("transpiling")} '${id}'`);
-		// the compiled JS code will not change if imports do, but the declarations can change based on imports' types changing, so only check imports when declarations are needed
-		return this.getCached(this.codeCache, id, snapshot, Boolean(this.options.declaration), transform);
+		// if !isolatedModules, compiled JS code can change if its imports do (e.g. enums). also, declarations can change based on imports as well
+		return this.getCached(this.codeCache, id, snapshot, Boolean(!this.options.isolatedModules || this.options.declaration), transform);
 	}
 
 	public getSyntacticDiagnostics(id: string, snapshot: tsTypes.IScriptSnapshot, check: () => tsTypes.Diagnostic[]): IDiagnostics[]
--- a/src/tscache.ts
+++ b/src/tscache.ts
@@ -213,7 +213,8 @@ export class TsCache
 	public getCompiled(id: string, snapshot: tsTypes.IScriptSnapshot, transform: () => ICode | undefined): ICode | undefined
 	{
 		this.context.info(`${blue("transpiling")} '${id}'`);
-		return this.getCached(this.codeCache, id, snapshot, false, transform);
+		// the compiled JS code will not change if imports do, but the declarations can change based on imports' types changing, so only check imports when declarations are needed
+		return this.getCached(this.codeCache, id, snapshot, Boolean(this.options.declaration), transform);
 	}
 
 	public getSyntacticDiagnostics(id: string, snapshot: tsTypes.IScriptSnapshot, check: () => tsTypes.Diagnostic[]): IDiagnostics[]


ezolenko:
I don't remember why it is this way, but I think your change is correct.

Now that I look at it closer, maybe we should always check imports regardless -- there might be cases where typescript generates different code based on what was imported. Maybe inlining enums into literals?

agilgur5:
> Maybe inlining enums into literals?

OH, that's actually the perfect edge case! I was trying to think of an edge-case where checking imports would be necessary for just compiled JS code (i.e. not declarations) and couldn't come up with one. But I did suspect that there were other edge-cases.

In the RCA I compared it to Babel:

> This seems to be because the JS code output shouldn't change if an import changes. That, I believe, is accurate. That's one of the reasons why Babel can just transpile TS code per-file (whereas type-checking requires knowledge of the whole "Program", i.e. the whole module graph).

And, of course, `enums` are one of those things that Babel _can't_ handle!

TS does have a check for this though, [`isolatedModules`](https://www.typescriptlang.org/tsconfig#isolatedModules).

So what I can do is `checkImports` if `!isolatedModules || declaration`. That's going to be the majority of cases, tbh, but it leaves in some potential for optimization.

Could make that a separate PR though, as checking `declaration` is still certainly needed

agilgur5:
added another commit to handle `!isolatedModules` case as well. modified the PR summary and title to match this as well.

think this is good to go now!

PR #366, opened 2022-06-23
feat: support `emitDeclarationOnly`

agilgur5:
## Summary

Support `emitDeclarationOnly` by, well, only emitting declarations and type-checking, and otherwise passing to the next plugin to transform TS
- Fixes #268

## Details

- when `emitDeclarationOnly` is set, only perform type-checking and emit declarations, don't transform TS
  - `result.code` actually doesn't exist when `emitDeclarationOnly` is set anyway
    - this caused a confusing situation for users who were trying to use Babel (with Babel plugins) on TS and only use rpt2 for declarations
      - instead of getting any JS code, they would just get empty chunks, bc `result.code` is `undefined`
        - that's kind of buggy, it should probably either be forced to `false` or do what we're doing now, which is likely more intuitive / intended
          - note that other Rollup TS plugins just force it to `false`, but I'm not sure that's the most optimal behavior
  - so now, instead of getting an empty chunk, rpt2 will just pass to the next plugin, allowing for other plugins on the chain to process TS
    - this opens up some new use cases, like using in tandem with Babel plugins (e.g. with [`babel-plugin-typescript-to-proptypes`](https://github.com/milesj/babel-plugin-typescript-to-proptypes)) or using for type-checking and declaration generation while using Vite / ESBuild for compilation

- add a new paragraph to "Declarations" docs about this new feature and what it does and some examples of how it could be used
  - note that these are unexplored / untested integrations, so point that out in the docs too
  - also modify a self-reference further up in the docs to use code backticks for `rollup-plugin-typescript2`
- while at it, also add a line about declaration maps
- and reformat the existing paragraph a bit to match the style and improve readability
  - one sentence per line, which is all on the same paragraph in Markdown anyway
  - add a `<br />` element to add a new-line _within_ the paragraph for better readability / spacing
    - Markdown supports this only with two trailing spaces, which is difficult to see and the trailing whitespace can be trimmed by editors, so prefer `<br />`
    
Can see my root cause analysis in https://github.com/ezolenko/rollup-plugin-typescript2/issues/268#issuecomment-1164003201

PR #371, opened 2022-06-29
test: add initial integration test suite

agilgur5:
## Summary

Creates a test harness/structure for integration tests and adds an initial integration test suite to it
- Gets a good way into the implementation of one of the longest-standing issues, lack of integration testing! Resolves one of the big leftover issues from https://github.com/ezolenko/rollup-plugin-typescript2/pull/311#issuecomment-1145551324
- Follow-up to #321

## Details

- some additional `ts-jest` config is needed to parse `index.ts`
  - since Rollup isn't used when importing files during testing, we need to enable `esModuleInterop`
    - the import of `findCacheDir` was immediately erroring without this (and maybe more, but `esModuleInterop` made the import work)
  - add `tsconfig.test.json` for this purpose
  - use `ts-jest` JSDoc types to get types for the `ts-jest` config as well

- add an `integration/` dir in the `__tests__/` dir for all integration tests
  - add a `fixtures/` dir in there as well for integration test fixtures
    - add a `tsconfig.json` for all fixtures to use
      - basically so that rpt2 source code is not `include`d, though also bc this may be good to customize for fixtures
    - add a simple fixture with "no-errors" that does an import and a type-only import

- add a simple integration test that just ensures that all files are part of the bundle
  - the bundled `index` file and each file's declaration and declaration map
  - and no _other_ files!
  - for Rollup, need to use paths relative to the root of the whole project
    - probably because that is `cwd` when running `jest` from root
    - will probably abstract out helpers here in the future as the test suite grows
  - 70% coverage of `index.ts` with just this one test!

- update `CONTRIBUTING.md` now that an integration test exists

**EDIT**: added more tests below!

- properly test the cache (92% coverage)
- test w/ and w/o declarations + declaration maps
- test case of plain JS that rpt2 skips and Rollup reads, which still makes it into the bundle
- test `allowJs` + `emitDeclarationOnly` from #366 
- test missing/non-existent `tsconfig` error
- test semantic and syntactic error
  - plus w/ `check: false` and `abortOnError: false` 

## Review Notes

Creating the harness itself was probably the biggest "mental" blocker for me (mostly in the number of unknowns to solve, but didn't end up being that complex). Now that this is created, I've already started adding more tests to it, namely for edge cases and error cases. Might add some to this PR pending when it gets merged, but otherwise will add those in additional PRs.
- ~Also I haven't quite figured out the error testing process for non-fatal errors yet, but I~ was able to _roughly_ get #345 tested with this harness as well, which is one reason why I felt more confident in moving that to "ready for review"
  - **EDIT**: have testing for warnings now 
- **EDIT**: added more tests in this PR as it has yet to be merged. and this creates a larger harness/structure for different types of tests too. I think only watch mode tests are missing now


agilgur5:
I've added a handful more integration tests now, so this is now a proper integration test suite, and not just a simple one-test harness anymore. Edited the title and description of the PR to match.

Had to bump test timeouts a couple times since integration tests take a good bit longer, especially when comparing two builds against each other (e.g. for cache / no cache).

A few more areas to add tests to:
- more `tsconfig` parsing / merging errors
- syntax errors / `emitSkipped`
- Rollup watch mode
- `allowJs` / the new `emitDeclarationOnly` support from #366 

But this PR is (still) ready to review/merge, those are extras and can be done in future PRs (same as the tests I added today on top of the simple harness from a week ago)

ezolenko:
Awesome!

Issue #47, opened 2017-12-14
support for pretty: true?


guidorice:
@ezolenko when my tsconfig has pretty: true, I can see
```
$ tsc

src/About/About.js(22,9): error TS1136: Property assignment expected.

22         ;doh
           ~
```
just curious if this plugin can report that too? Fair enough it does give the row,column info like this:

> [!] (rpt2 plugin) Error: ...path.../src/About/About.js (22,9): syntax error TS1136 Property assignment expected.
src/About/About.js

But it would be super awesome to show what tsc shows for the context. I tried adjusting the verbosity but that did not seem to do it.

ezolenko:
Yeah, that's a good idea. Try with latest commit, should follow `pretty` option now.

guidorice:
Thanks much! very awesome work :)

Issue #103, opened 2018-08-08
Duplicate error message (esp. when `pretty: true`)


kitsonk:
#### What happens and why it is wrong

Duplicate semantic errors get logged to the console, which causes confusion.  For example:

```
./node_modules/rollup/bin/rollup -c ../../rollup.config.js -i ../../js/main.ts -o gen/bundle/main.js --silent
[!] (rpt2 plugin) Error: ../../js/main.ts:8:1 - error TS2304: Cannot find name 'confounded'.

8 confounded;
  ~~~~~~~~~~

../../js/main.ts
Error: ../../js/main.ts:8:1 - error TS2304: Cannot find name 'confounded'.

8 confounded;
  ~~~~~~~~~~

    at error (/Users/kkelly/github/deno/third_party/node_modules/rollup/dist/rollup.js:3365:15)
    at Object.error (/Users/kkelly/github/deno/third_party/node_modules/rollup/dist/rollup.js:20727:13)
    at RollupContext.error (/Users/kkelly/github/deno/third_party/node_modules/rollup-plugin-typescript2/src/rollupcontext.ts:36:18)
    at lodash_2 (/Users/kkelly/github/deno/third_party/node_modules/rollup-plugin-typescript2/src/print-diagnostics.ts:37:10)
    at arrayEach (/Users/kkelly/github/deno/third_party/node_modules/rollup-plugin-typescript2/node_modules/lodash/lodash.js:516:11)
    at forEach (/Users/kkelly/github/deno/third_party/node_modules/rollup-plugin-typescript2/node_modules/lodash/lodash.js:9342:14)
    at printDiagnostics (/Users/kkelly/github/deno/third_party/node_modules/rollup-plugin-typescript2/src/print-diagnostics.ts:9:2)
    at Object.transform (/Users/kkelly/github/deno/third_party/node_modules/rollup-plugin-typescript2/src/index.ts:238:5)
    at /Users/kkelly/github/deno/third_party/node_modules/rollup/dist/rollup.js:20750:41
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

I am not entirely sure how much this is `rollup-plugin-typescript2` and how much of this is just the way rollup behaves, but I couldn't find an answer either way.

#### Versions
- typescript: 3.0.1
- rollup: 0.63.2
- rollup-plugin-typescript2: 0.16.1

vineetit1991:
any input on this ? I am also facing this issue

[...]

PR #372, opened 2022-06-29
fix(diagnostics): `pretty` defaults to `true` in TS 2.9+

agilgur5:
## Summary

Changes the default configuration of `pretty`, so now pretty-printing is the default, as it is for `tsc` and in the TSConfig Reference.

## Details

- per https://www.typescriptlang.org/tsconfig#pretty
  - TS 2.9 [changed the default](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#--pretty-output-by-default) to `true`
     - so this is a legacy remnant of older versions of TS
- this is why `tsc` by default pretty prints, even when `pretty` is unset
  - only if it is _explicitly_ set to `false` does it not do that

- so change rpt2's diagnostic printing to be `pretty` by default as well
  - I think this is a pretty _big_ DX improvement, to be honest
  - I always thought something was up with the error reporting, but saw that there was code for `pretty`, so didn't quite connect the dots until now
    - that while there was code for it, the default was wrong, and so unless I set `pretty: true`, I wasn't getting easier to read printing
      - and given that `pretty: true` is the default, that's a redundant / unnecessary option to set in `tsconfig` that I normally wouldn't ever re-set

## Review Notes

Only thing I'm not sure of is non-tty environments, that I didn't even think of until I read the TS 2.9 release notes (as linked above). Not exactly sure if TS will handle those automatically, or if we need to add a check for that as well... 🤔 

It's also a bit hard to check as well... though first priority should be local dev in either case.

## References

- Support for pretty-printing diagnostics was originally added in #47 , which I believe pre-dates TS 2.9

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -57,7 +57,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 	const typecheckFile = (id: string, snapshot: tsTypes.IScriptSnapshot, tcContext: IContext) =>
 	{
 			const diagnostics = getDiagnostics(id, snapshot);
-			printDiagnostics(tcContext, diagnostics, parsedConfig.options.pretty === true);
+			printDiagnostics(tcContext, diagnostics, parsedConfig.options.pretty !== false);
 
 			if (diagnostics.length > 0)
 				noErrors = false;
@@ -131,7 +131,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 			// printing compiler option errors
 			if (pluginOptions.check) {
 				const diagnostics = convertDiagnostic("options", service.getCompilerOptionsDiagnostics());
-				printDiagnostics(context, diagnostics, parsedConfig.options.pretty === true);
+				printDiagnostics(context, diagnostics, parsedConfig.options.pretty !== false);
 				if (diagnostics.length > 0)
 					noErrors = false;
 			}
--- a/src/parse-tsconfig.ts
+++ b/src/parse-tsconfig.ts
@@ -20,7 +20,7 @@ export function parseTsConfig(context: IContext, pluginOptions: IOptions)
 	let loadedConfig: any = {};
 	let baseDir = pluginOptions.cwd;
 	let configFileName;
-	let pretty = false;
+	let pretty = true;
 	if (fileName)
 	{
 		const text = tsModule.sys.readFile(fileName);
--- a/src/print-diagnostics.ts
+++ b/src/print-diagnostics.ts
@@ -4,7 +4,7 @@ import { tsModule } from "./tsproxy";
 import { IContext } from "./context";
 import { IDiagnostics } from "./tscache";
 
-export function printDiagnostics(context: IContext, diagnostics: IDiagnostics[], pretty: boolean): void
+export function printDiagnostics(context: IContext, diagnostics: IDiagnostics[], pretty = true): void
 {
 	diagnostics.forEach((diagnostic) =>
 	{


agilgur5:
ah sh*t, while testing #345, it seems like this could give duplicate errors. let me look into this before merging pl0x

agilgur5:
Ok that's not a duplicate error... or rather there's already a duplicate error:

w/o `pretty`:
```
./src/index.ts → ./dist/index.ts...
[!] (plugin rpt2) Error: /project-dir/src/difference.ts(1,36): semantic error TS6133: 'a' is declared but its value is never read.
src/difference.ts
Error: /project-dir/src/difference.ts(1,36): semantic error TS6133: 'a' is declared but its value is never read.
    at error (/project-dir/node_modules/rollup/dist/shared/rollup.js:198:30)
    at throwPluginError (/project-dir/node_modules/rollup/dist/shared/rollup.js:21936:12)
    at Object.error (/project-dir/node_modules/rollup/dist/shared/rollup.js:22659:20)
    at Object.error (/project-dir/node_modules/rollup/dist/shared/rollup.js:22113:42)
    at RollupContext.error (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:20259:30)
    at /project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29527:23
    at Array.forEach (<anonymous>)
    at printDiagnostics (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29500:17)
    at Object.transform (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29785:17)
    at /project-dir/node_modules/rollup/dist/shared/rollup.js:22868:37
```

w/ `pretty`:
```
./src/index.ts → ./dist/index.ts...
[!] (plugin rpt2) Error: src/difference.ts:1:36 - error TS6133: 'a' is declared but its value is never read.

1 export default function difference(a: number, b: number) {
                                     ~
src/difference.ts:1:47 - error TS6133: 'b' is declared but its value is never read.

1 export default function difference(a: number, b: number) {
                                                ~

src/difference.ts
Error: src/difference.ts:1:36 - error TS6133: 'a' is declared but its value is never read.

1 export default function difference(a: number, b: number) {
                                     ~
src/difference.ts:1:47 - error TS6133: 'b' is declared but its value is never read.

1 export default function difference(a: number, b: number) {
                                                ~

    at error (/project-dir/node_modules/rollup/dist/shared/rollup.js:198:30)
    at throwPluginError (/project-dir/node_modules/rollup/dist/shared/rollup.js:21936:12)
    at Object.error (/project-dir/node_modules/rollup/dist/shared/rollup.js:22659:20)
    at Object.error (/project-dir/node_modules/rollup/dist/shared/rollup.js:22113:42)
    at RollupContext.error (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:20259:30)
    at /project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29524:19
    at Array.forEach (<anonymous>)
    at printDiagnostics (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29500:17)
    at Object.transform (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29785:17)
    at /project-dir/node_modules/rollup/dist/shared/rollup.js:22868:37
```

The error that causes Rollup to bail gets printed out once, then the `id` of the file, then the error again (in gray) plus the stack trace. This might be Rollup behavior? Just with the `pretty` errors it's much more noticeable

agilgur5:
Oh, I literally just reproduced #103 😅😮 
It previously didn't have a repro, but I guess it's just more visible when using `pretty`

agilgur5:
So setting `abortOnError: false` makes it no longer duplicate:

w/o `pretty`:
```
./src/index.ts → ./dist/index.ts...
(!) Plugin rpt2: /project-dir/src/difference.ts(1,36): semantic error TS6133: 'a' is declared but its value is never read.
src/difference.ts
(!) Plugin rpt2: /project-dir/src/difference.ts(1,47): semantic error TS6133: 'b' is declared but its value is never read.
src/difference.ts
created ./dist/index.ts in 575ms
```

w/ `pretty`:
```
./src/index.ts → ./dist/index.ts...
(!) Plugin rpt2: src/difference.ts:1:36 - error TS6133: 'a' is declared but its value is never read.

1 export default function difference(a: number, b: number) {
                                     ~
src/difference.ts:1:47 - error TS6133: 'b' is declared but its value is never read.

1 export default function difference(a: number, b: number) {
                                                ~

src/difference.ts
src/difference.ts
created ./dist/index.ts in 586ms
```

So if I'm understanding this correctly, I think Rollup re-throws an error from a plugin or something, causing this behavior.
I.e. from Rollup's perspective, there's an error printed at "plugin level" and then at "Rollup level" as well because it was fatal.

Changing the `Context` code to use `throw` instead of `context.error` results in the same behavior, so that doesn't fix it 😕  
Perhaps catching in `buildEnd` might?

agilgur5:
Ok, well I think I've figured it out, and this really is probably more a Rollup bug, but we can probably(?) workaround it.

What Rollup is printing out at the end is the error [`stack`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack), which happens to contain the error itself, as far as I can tell.

So, with this code in `buildEnd`:
```ts
this.error({ message: "rpt2 error (see below)", stack: err.stack });
```

We get this log instead (no duplicate):
```
./src/index.ts → ./dist/index.ts...
[!] (plugin rpt2) Error: rpt2 error (see below)
Error: src/difference.ts:1:36 - error TS6133: 'a' is declared but its value is never read.

1 export default function difference(a: number, b: number) {
                                     ~
src/difference.ts:1:47 - error TS6133: 'b' is declared but its value is never read.

1 export default function difference(a: number, b: number) {
                                                ~

    at error (/project-dir/node_modules/rollup/dist/shared/rollup.js:198:30)
    at throwPluginError (/project-dir/node_modules/rollup/dist/shared/rollup.js:21936:12)
    at Object.error (/project-dir/node_modules/rollup/dist/shared/rollup.js:22659:20)
    at Object.error (/project-dir/node_modules/rollup/dist/shared/rollup.js:22113:42)
    at RollupContext.error (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:20259:30)
    at /project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29524:19
    at Array.forEach (<anonymous>)
    at printDiagnostics (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29500:17)
    at Object.transform (/project-dir/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29791:17)
    at /project-dir/node_modules/rollup/dist/shared/rollup.js:22868:37
 ```
 So it prints `rpt2 error (see below)` and then the stack trace, which has the full, `pretty` message embedded within it.
 
`Error.stack` is non-standardized; so the only issue is that this could potentially vary by runtime... We're only targeting Node, that being said, and this should be the same in all LTS Node versions (per the compat table in the link above, been around since Node 0.10 at least)

agilgur5:
Yea pretty sure this is a Rollup bug; [this line](https://github.com/rollup/rollup/blob/56c1514c7b0d0c985685ed497bdc6ce33b023383/cli/logging.ts#L14) prints `toString` and then later it _also_ prints [`stack`](https://github.com/rollup/rollup/blob/56c1514c7b0d0c985685ed497bdc6ce33b023383/cli/logging.ts#L31), which will _also_ print the error message -- resulting in a duplicate -- and then print the stack trace. 😕

agilgur5:
Added a fix for the duplicate error messages / #103 in #373 . That should be merged along with this PR as this PR makes it more noticeable, but that bug already exists without `pretty` as well, so they're technically independent.

PR #374, opened 2022-07-02
clean: remove backward-compat checks for old Rollup versions

agilgur5:
## Summary

Clean up some old backward-compat checks that are no longer necessary for the minimum version of Rollup that rpt2 requires
- Follow-up to https://github.com/ezolenko/rollup-plugin-typescript2/pull/345#issuecomment-1170542354, where I realized several of these checks were legacy and no longer needed

## Details
- we require Rollup `>=1.26.3` in the peerDeps and the README and have for years now
  - (and Rollup is currently at `2.75.7`, so this is a pretty low target)
  - as such, we can remove various checks that are legacy backward-compat remnants for old Rollup versions:
    - `this.meta` was added to `options` in `1.1.0`: https://github.com/rollup/rollup/blob/master/CHANGELOG.md#110
    - `this.addWatchFile` was added at least in `1.0.0`: https://github.com/rollup/rollup/blob/master/CHANGELOG.md#100
    - `this.error` and `this.warn` were added to `transform` in `0.41.0`: https://github.com/rollup/rollup/blob/master/CHANGELOG.md#0410

- this simplifies some of the code a decent bit, `RollupContext` in particular
  - see also the usage of `buildEnd` in #345

- modify tests to account for these changes; basically just simplify them
  - and remove the check against private internals of the class, as they're private
    - guess I forgot to remove this when I was refactoring the test code in #321?

## Future Work

- More follow-ups to #328 to remove more `lodash` usage
  - The rest of the `_.isFunction` usage can be replaced with a check for the opposite condition, i.e. `typeof message === "string" ? message : message()`
  - `_.merge`, `_.isEqual`, and `_.compact` would be the only ones leftover after

### DIFF
--- a/__tests__/rollupcontext.spec.ts
+++ b/__tests__/rollupcontext.spec.ts
@@ -12,9 +12,7 @@ import { RollupContext } from "../src/rollupcontext";
 test("RollupContext", () => {
 	const data = {};
 	const stubbedContext = makeStubbedContext(data);
-
 	const context = new RollupContext(5, false, stubbedContext);
-	expect(Object.keys(context)).toEqual(["verbosity", "bail", "context", "prefix", "hasContext"]);
 
 	context.warn("test");
 	expect((data as any).warn).toEqual("test");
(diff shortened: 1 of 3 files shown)


ezolenko:
btw, there is no reason not to bump minimum rollup version if we need to use anything new, I expect most people either update whole build toolchain, or freeze the lot. Well, all sane people would anyway... :)

agilgur5:
> btw, there is no reason not to bump minimum rollup version if we need to use anything new

Yea I figure, given how old the current min is, but that also would technically be a breaking change, so also wouldn't want to do so without good reason. So far, I haven't encountered anything that needed new APIs though (which is surprising, because as a Rollup user, it seemed like I had to update quite a lot for [Rollup 2](https://github.com/jaredpalmer/tsdx/pull/889), but now, as a plugin developer, I'm not seeing the same need to upgrade 🤷  )

PR #381, opened 2022-07-10
fix: add back `noErrors = false` to `emitSkipped`

agilgur5:
## Summary

Make `noErrors = false` unconditional again in `emitSkipped`
- Follow-up to https://github.com/ezolenko/rollup-plugin-typescript2/pull/344#discussion_r914336553

## Details

- apparently `emitSkipped` can indeed happen without there being a semantic or syntactic error
  - so add back the previous unconditional behavior of `noErrors`

- still not clear what exactly triggers `emitSkipped` though, see my TS issue https://github.com/microsoft/TypeScript/issues/49790

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -198,6 +198,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 
 				if (output.emitSkipped)
 				{
+					noErrors = false;
 					// always checking on fatal errors, even if options.check is set to false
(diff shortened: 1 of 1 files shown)


agilgur5:
Oh... well I didn't quite realize this (was refactoring/simplifying this code a bit after all), but this is totally redundant because `this.error` bails out anyway, so the single log statement that `noErrors` is used for will never be reached anyway 😐

PR #383, opened 2022-07-10
fix: missing `tsconfig` error shouldn't say `undefined`

agilgur5:
## Summary

Improve the error message when a `tsconfig` path doesn't have a file. The error message should say the path passed in, not `'undefined'`
- Fixes a bug I found while writing integration tests in #371 

## Details

- per the conditional above this line, `file` is falsey, so printing it doesn't make sense
  - per same conditional though, `pluginOptions.tsconfig` exists, so we can print that

- fixes a test `TODO`/`FIXME` that had to workaround this bug as well


agilgur5:
Tests are failing due to #377 ; the test for this change passes

Issue #68, opened 2018-03-21
Disable cache


lastmjs:
Could we possibly provide a way to disable the cache? I have my own caching built into the tool that I'm using this plugin to build, and I don't want the `.rpt2_cache` folder lying around afterwards, nor do I necessarily want the override of writing to the cache

ezolenko:
What level are you caching at? Is it going to bypass typescript only or whole bundle? I'm caching both compiled code and diagnostic data, so most of the time unchanged files are not passed to typescript at all, while still repeating errors found on the previous run.

Otherwise a flag to disable caching should be easy enough to do.

wessberg:
I would argue that the “clean” option should disable the cache. If the cache is cleaned for each rebuild, why cache at all? 😊

ezolenko:
That's a good point, better than adding a new option I was originally planing to do :)

ezolenko:
@lastmjs could you check if master works for you with `clean: true`?

PR #387, opened 2022-07-20
docs: clarify `clean: true` and default `include`

agilgur5:
## Summary

Minor docs clarifications, fixes, and style consistency
- Follow-up to #375

## Details

- the docs for `clean` don't seem to have been updated after f15cb84dcc99a0bd20f3afce101c0991683010b6 (#68), which slightly changed how it works
- also clarify that "wipes out cache" means it deletes _all_ previous caches by saying "wipes any existing cache" instead

- fix the `include` default explanation to mention `.tsx` files as well, since the regex includes those
  - also add code backticks and capitalize TypeScript etc (guess I missed this option in my previous docs PRs? or I intentionally left it as a separate change a while ago...)

- also add code backticks around `node_modules` in `cacheRoot` as well while at it

- use non-rendered newlines between sentences for markdown style consistency

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,15 +144,17 @@ See [#108](https://github.com/ezolenko/rollup-plugin-typescript2/issues/108)
 
 * `clean`: false
 
-	Set to true for clean build (wipes out cache on every build).
+	Set to true to disable the cache and do a clean build.
+	This also wipes any existing cache.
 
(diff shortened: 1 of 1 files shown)


agilgur5:
Windows CI failures are due to #385 (we don't have any docs tests at this time, so docs-only changes can't result in a CI failure anyway)

PR #393, opened 2022-07-22
refactor(cache): further condense `walkTree`

agilgur5:
## Summary

Condense `walkTree` further to just 4 LoC
- Follow-up to #359 I guess

## Details

- `forEach` returns `void`, so we can just `return` it and condense the `if` a bit

### DIFF
--- a/src/tscache.ts
+++ b/src/tscache.ts
@@ -188,13 +188,9 @@ export class TsCache
 	public walkTree(cb: (id: string) => void | false): void
 	{
 		if (alg.isAcyclic(this.dependencyTree))
-		{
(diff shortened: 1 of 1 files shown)


ezolenko:
Feels weird to return void, but I guess TS will warn if return types ever get out of sync 🤷

agilgur5:
Yea it's a stylistic thing; I thought it seemed odd in the context of this codebase but I've written code like this in other codebases where it doesn't seem as "out of place".
Changing some of the style in this and #402 should make it feel more comfortable.

Should probably get `prettier` (or another formatter) into this codebase at some point to make everything more consistent (and switch to non-deprecated ESLint too 😅)

PR #396, opened 2022-07-22
refactor: combine two context files into one

agilgur5:
## Summary

Merge the two context files into a single file as they're _intended_ to be as similar as possible. They're also _not_ meant to be more extensible to create other contexts (unlike say, the cache interfaces etc), these are the only two contexts intended.

IMO, the slight, small distinction between them would've been easier to understand/catch on to if I saw them together like this than in separate files.

## Details

- they're virtually identical, so combine them instead of keeping them separate
  - changes to one should probably be made to both
  - still a < 100 LoC file

- refactor out `_.isFunction` with a simple `getText` function instead
  - checks the opposite
  - one more `lodash` removal! (follow-up to #328)

- add docstrings about when to use the two contexts

## Future Work

- I'd like to do something similar with the `diagnostics` files and the `tsconfig` files, as they're also only used in the same context and not in other contexts. Planning to add more unit tests first though


agilgur5:
Very likely that `ConsoleContext` will be removed per my realization in https://github.com/ezolenko/rollup-plugin-typescript2/pull/345#issuecomment-1200277838, but this PR will also make it easier to remove it as all the code is now in this one file and all the imports are from this one file

PR #403, opened 2022-07-29
test: ensure `declarationMap` sources are correct 

agilgur5:
~**NOTE**: this is built on top of #386 as the `testDir` changes there are required to have a correct `outDir`. As such, I've marked this PR as "Draft" until that PR is merged.~
Rebased on top and marked as ready for review.

## Summary

Add a test for `declarationMap` `sources` to make sure they're correct after #221 / #204
- Follow-up to https://github.com/ezolenko/rollup-plugin-typescript2/pull/371#issuecomment-1174339120

## Details

- since we set a placeholder `outDir` initially, TS sets declaration map sources to that placeholder dir
  - then, after #221, we remap that to the correct dir that Rollup outputs to
    - this test checks that the remap happens and is correct

### DIFF
--- a/__tests__/integration/no-errors.spec.ts
+++ b/__tests__/integration/no-errors.spec.ts
@@ -1,6 +1,8 @@
 import { jest, afterAll, test, expect } from "@jest/globals";
 import * as path from "path";
 import * as fs from "fs-extra";
+import { OutputAsset } from "rollup";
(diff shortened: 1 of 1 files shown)


agilgur5:
CI failed on an integration test suite timeout, but this PR itself shouldn't have increased the time by more than some milliseconds, so think it makes sense to just re-run CI in this case.

PR #406, opened 2022-07-29
fix: handle all type-only imports by piping TS imports

agilgur5:
~**NOTE**: this is built on top of #403 as it changes its code a bit. #403 is itself built on top of #386. As such, I've marked this PR as "Draft" until those PRs are merged.~
Rebased on top and marked as ready for review. (surprisingly clean rebase)

Also this might be the first time this repo has more open PRs than open issues 😮 

## Summary

Completely handle the long-standing issue of type-only imports
- Uses the solution I proposed in my root cause analysis in https://github.com/ezolenko/rollup-plugin-typescript2/issues/298#issuecomment-1146658442
- Fixes #7 as all type-only imports go through the `resolveId` hook now, which adds them to the cache (calls `cache().setDependency`)
- Fixes #211
  - And its various duplicates: fixes #24, fixes #106, fixes #409
- Fully fixes the most common issue that I mentioned in https://github.com/ezolenko/rollup-plugin-typescript2/pull/311#issuecomment-1115491161, type-only imports
- Related to #345
  - _That_ PR fixed #298 and the latter half of #254 for their specific use-cases as they happen to use `include` globs, but _this_ PR fixes it in all other scenarios as well (namely when using [`files`](https://www.typescriptlang.org/tsconfig#files) instead of `include` globs)

## Details

- `result.references` is populated by `ts.preProcessFile`; i.e. this is TS discovering all imports, instead of Rollup
  - TS's imports include type-only files as TS understands those (whereas they aren't emitted in the JS for Rollup to see, since, well, they produce no JS)
  - so we can pipe all these through Rollup's `this.resolve` and `this.load` to make them go through Rollup's `resolveId` -> `load` -> `transform` hooks
    - this makes sure that other plugins on the chain get to resolve/transform them as well
    - and it makes sure that we run the same code that we run on all other files on type-only ones too
      - for instance: adding declarations, type-checking, setting them as deps in the cache graph, etc
      - yay recursion!
        - also add check for circular references b/c of this recursion (which Rollup docs confirm is necessary, per in-line comment)
    - and Rollup ensures that there is no perf penalty if a regular file is processed this way either, as it won't save the hook results when it appears in JS (i.e. Rollup's module graph)
      - we are checking more files though, so that in and of itself means potential slowdown for better correctness

- add a test for this that uses a `tsconfig` `files` array, ensuring that the `include` workaround won't cover these type-only files
  - this test fails without the new code added to `index` in this commit
[...]

PR #410, opened 2022-08-17
revert: back to using `vue.d.ts` instead of `.d.ts` for Vue declarations

agilgur5:
This reverts commit e145d0baebbbef1eec691b99d1ce0dce1e49f38c / PR #336
**NOTE**: this may be considered breaking, so we may want to release this in a breaking release as such. I've left this PR as draft as such. That being said, this could also be considered as just a bugfix as the original PR was made in error and caused a bug.

## Summary

Revert back to using `.vue.d.ts` for Vue declarations

## Details

- Per follow-up discussion on [the original issue](https://github.com/ezolenko/rollup-plugin-typescript2/issues/224#issuecomment-1193350666) and [to-be-reverted PR](https://github.com/ezolenko/rollup-plugin-typescript2/pull/336#issuecomment-1179476746), it seems that the request to use `.d.ts` instead of `.vue.d.ts` was made in error
  - `.d.ts` seems to only be necessary if Vue users were importing `.vue` SFCs without extensions ("extensionless")
    - i.e. `import MyComponent from "./MyComponent"` instead of `import MyComponent from "./MyComponent.vue"`
  - and "extensionless" imports are no longer supported by the Vue team (but used to be)
  - requiring extensionless imports also breaks imports when extensions _are_ used
  - so these are not necessarily compatible with each other, but the Vue team support strongly suggests that `.vue.d.ts` would be the proper way forward

## Credits

Thanks to @andrew0 for finding this problem and providing evidence 🙂

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -345,7 +345,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 
 				let fileName = entry.name;
 				if (fileName.includes("?")) // HACK for rollup-plugin-vue, it creates virtual modules in form 'file.vue?rollup-plugin-vue=script.ts'
-					fileName = fileName.split(".vue?", 1) + extension;
+					fileName = fileName.split("?", 1) + extension;
 
 				// If 'useTsconfigDeclarationDir' is in plugin options, directly write to 'declarationDir'.
 				// This may not be under Rollup's output directory, and thus can't be emitted as an asset.


ezolenko:
Yeah, we should support official way of importing, so looks good.

Do we want to release this sooner than later? Any other PRs you want to include (or exclude)?

agilgur5:
> Do we want to release this sooner than later? Any other PRs you want to include (or exclude)?

So depends if we're considering this as a breaking change or not.
If it's a patch, then I think this can go out with all the other already merged changes into `0.32.2`.
If we're considering this as breaking, then it probably makes sense to release along with #406 . So then I'd think to release `0.32.2` with the already merged changes and this is and #406 as part of `0.33.0`.

Exclude-wise, I was just thinking of leaving #406 until either `0.32.3` or `0.33.0` (pending patch/breaking classification), and releasing `0.32.2` on its own without #406. Just because #406 is a _big_ change and there's enough changes already merged in for a release.

ezolenko:
I think we should do 0.33 then without #406, and then 0.34 with that when it is ready

ezolenko:
@agilgur5 0.33.0 is out, maybe check if release notes make sense and edit them if you want anything changed

agilgur5:
> I think we should do 0.33 then without #406, and then 0.34 with that when it is ready

Ok, up to you 👍   Your releases consistently err on the side of breaking (vs. patch), so that's maybe more expected for rpt2 users too, in that sense.

> @agilgur5 0.33.0 is out, maybe check if release notes make sense and edit them if you want anything changed

Thanks for (mostly) copying my release notes style!
I made a few adjustments, for reference: 
- put "Optimizations", "Docs", and "GitHub" sections outside of the "Other"/"Internal" section as they are user-facing and so may be good to highlight
- put this PR, which begins with `revert: ` into the "Bugfixes" section and put the `dx: ` PRs into the "Docs" section (that's kind of the closest thing to it)
- highlighted a few of the "Internal" changes before the spoiler too -- think users would be happy to see all the testing additions at least!
- hid a few fixes etc under a spoiler where they were not user-facing or very minor in nature
- added explanations to several of the bugfixes where the PR title would not necessarily be understood by users

I also saw that the `dist/` build was missing in the release initially, then saw that you re-tagged it. Looks like NPM has the build already, so seems all good minus some tag confusion 👍

ezolenko:
I basically took autogenerated release notes with your commit messages and rearranged them to hide internal stuff, so those are literally your release notes. :)

Yeah, forgot to push local branch with build commit before tagging on github :/

agilgur5:
> I basically took autogenerated release notes with your commit messages and rearranged them to hide internal stuff, so those are literally your release notes. :)

Yep that's the basic gist of the structure I use 👍   plus a few manual things where explanations could be helpful

> Yeah, forgot to push local branch with build commit before tagging on github :/

Yea I've done that multiple times before too 😅 

If you're interested, we can also automate the release process with GH Actions too

PR #413, opened 2022-08-24
refactor: combine `check-tsconfig` with `parse-tsconfig`

agilgur5:
~**NOTE**: this is built on top of #404 as it uses the `context` changes made there. As such, I've marked this PR as "Draft" until #404 is merged.~
Rebased on top and marked as ready for review.

## Summary

Combine the two `tsconfig` files into one as the split is not currently necessary and adds unnecessary complexity
- Follow-up to #396

## Details

- `checkTsConfig` is literally only used by `parseTsConfig`
  - I first just moved the function over, but it's a two-liner, so thought it made more sense to just in-line it

- merge the unit tests as well
  - we already check the non-error case in `parse-tsconfig.spec`, so only add the error case

- `check-tsconfig` was longer in the past and more files were being created then for separation, so this may have made more sense then, but now it is unnecessary
  - c.f. 733207791800ffe15672554d2a6337c73c65cf74


agilgur5:
This is gonna merge conflict with #404 -- please merge that first as it's a complicated one that a bunch of other code depends on

agilgur5:
oh woops, these seem to merge cleanly actually, so this should be fine? I wrote this a bit ago with the queue, so I probably anticipated this already and forgot

Issue #86, opened 2018-06-06
Switch to shallow merge in tsconfigOverride where typescript does shallow merge


ezolenko:
#### What happens and why it is wrong

typescript is using shallow merge on some elements (paths, possibly includes etc). Need to make plugin behave the same way to reduce surprises.

agilgur5:
Per the TS docs on [`extends`](https://www.typescriptlang.org/tsconfig#extends)  -- which we should treat as a "source of truth" -- the fields that should be shallow merged (replaced, rather) are `files`, `include`, and `exclude`. 

Basically all the top-level arrays.
`paths` isn't mentioned. It's possible that's just missing docs though, but I haven't checked the TS API to confirm which it does.
Also `references` specifically _isn't_ inherited, so not sure if we want to have the same behavior for that one or not.

PR #416, opened 2022-08-30
refactor: consolidate `diagnostics` funcs into single file 

agilgur5:
## Summary

- Move `diagnostics` funcs into a single file named `diagnostics.ts`
  - Follow-up to #396 (and #413)
    - This should be the last of such "consolidation" PRs, as these 3 files were the only ones where I thought the separation was more confusing than helpful, and where consolidation simplified things, IMO
- Add a unit test for `convertDiagnostic`
- Tiny bugfix `\n` -> `formatHost.getNewLine()`
  - This doesn't seem to have ever affected anyone; I'm guessing either Node or people's terminals auto-translated the newlines b/t OSes or something? So this is more of a semantic bugfix in that sense

## Details

refactor:
- move `IDiagnostic` and `convertDiagnostic` from `tscache` to `print-diagnostics`, then rename it to `diagnostics.ts`
  - the diagnostic funcs being in `tscache` always felt like a strange place for them
    - especially when `parse-tsconfig` or `print-diagnostics` would import them from `tscache`, which sounded like an unrelated file

- may want to move `diagnostics-format-host` into this file as well, as it is _only_ used in this file
  - leaving as is for now, limiting this change to a smaller one

test: add unit test for `convertDiagnostic`
- this was previously only covered in integration tests
  - since unit tests don't currently touch `index` or `tscache`, and `convertDiagnostic` was previously in `tscache`
    - another reason why consolidating these functions into one file made sense
    
 fix(diagnostics): use `formatHost.getNewLine()` instead of `\n`
- since new lines are host OS specific

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,5 +73,5 @@ A useful resource as you dive deeper are the [unit tests](__tests__/). They're g
 1. At this point, you may be ready to read the more complicated bits of [`index`](src/index.ts) in detail and see how it interacts with the other modules.
     - The [integration tests](__tests__/integration/) could be useful to review at this point as well.
 1. Once you're pretty familiar with `index`, you can dive into some of the cache code in [`tscache`](src/tscache.ts) and [`rollingcache`](src/rollingcache.ts).
-1. And finally, you can see some of the Rollup logging nuances in [`context`](src/context.ts) and then the TS logging nuances in [`print-diagnostics`](src/print-diagnostics.ts), and [`diagnostics-format-host`](src/diagnostics-format-host.ts)
+1. And finally, you can see some of the Rollup logging nuances in [`context`](src/context.ts) and then the TS logging nuances in [`diagnostics`](src/diagnostics.ts), and [`diagnostics-format-host`](src/diagnostics-format-host.ts)
     - While these are necessary to the implementation, they are fairly ancillary to understanding and working with the codebase.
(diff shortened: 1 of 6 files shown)


agilgur5:
> * Tiny bugfix `\n` -> `formatHost.getNewLine()`
>   * This doesn't seem to have ever affected anyone; I'm guessing either Node or people's terminals auto-translated the newlines b/t OSes or something? So this is more of a semantic bugfix in that sense

Well this was pretty ironic -- the unit test I added actually fails on Windows without this fix (and a test adjustment). Given that I'm on Mac, I literally didn't see this until it hit CI after I wrote up the PR. Turns out the test confirms this behavior!

agilgur5:
@ezolenko friendly ping in case you just hadn't seen this PR (since you merged some others, but not this one)

I also think we should be good to release `0.34.0`, mainly containing #406 as that's a big enough change as is.
`0.33.0` has already been out a few weeks without issues, so don't think we need to worry about releasing two minors shortly after each other anymore.

There's a couple more things that I'm working on, but those will probably take some time (not spending as much time on rpt2 anymore since I've fixed the vast majority of issues! 🙂), so don't want to hold back releasing #406 because of that, since it's a big fix for what is probably the single most common issue in this repo.
Also the stuff I'm working on is mostly bugfixes and optimizations for cache/watch mode (i.e. follow-ups to #386), which can go out whenever as patch releases. #228 and #86 are the only real leftover bugs, and fixes to those probably should be in a separate minor than #406 anyway, IMO. Those might also be a few weeks away too (or more or less -- don't have much of a timeline on those in general), as they're not high priority in my "OSS todo list".

ezolenko:
@agilgur5 0.34.0 is out on npm now, pls check if release notes on github make sense. Great work, thanks!

PR #420, opened 2022-09-10
feat(cache): set options.clean as 'true' by default

oscard0m:

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -86,7 +86,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 		{
 			check: true,
 			verbosity: VerbosityLevel.Warning,
-			clean: false,
+			clean: true,
 			cacheRoot: findCacheDir({ name: "rollup-plugin-typescript2" }),
 			include: ["*.ts+(|x)", "**/*.ts+(|x)"],
 			exclude: ["*.d.ts", "**/*.d.ts"],
--- a/src/tscache.ts
+++ b/src/tscache.ts
@@ -118,6 +118,7 @@ export class TsCache
 
 		if (noCache)
 		{
+			this.context.warn(yellow("Cleaning cache... If you want to use caching, please set `clean` option to false."));
 			this.clean();
 			return;
 		}


agilgur5 (changes requested):
> * Where is the `extraCacheKeys` option declared? I don't find it. ([_originally posted here](https://github.com/ezolenko/rollup-plugin-typescript2/issues/228#issuecomment-1181971835))

It doesn't exist yet -- this PR would be creating it. It would be in the same `index.ts` block where you changed `clean` currently.

> * [ ]  Set `clean` option to `true` by default

It should _not_ be `true` by default -- see below

> * Does this mean I need to do some logic to decide the default value for `clean` option depending on these 2 other options?

Yes. Here's a quick logic flow:
1. `clean` is `false` by default
2. If `sourceMapCallback` or `transformers` are passed in, then it should default to `true` instead 
3. If `extraCacheKeys` is passed in, then it should default back to `false` again

`2` should also log out, telling the user to use `extraCacheKeys` if they want to re-enable the cache.

oscard0m:
Applied more changes to the draft.

Now it's not clear to me which part of the code (and how) the `extraCacheKeys` should be treated.

src/ioptions.ts
```diff
@@ -30,4 +30,5 @@ export interface IOptions
 	tsconfigDefaults: any;
 	sourceMapCallback: (id: string, map: string) => void;
 	objectHashIgnoreUnknownHack: boolean;
+	extraCacheKeys: string[]
```
agilgur5, line 33:
```suggestion
	extraCacheKeys: any
```

This can contain literally any variables etc to add to the hash. It can be a single var, an object, an array, etc. Doesn't have to be only strings either.

src/tscache.ts
```diff
@@ -118,6 +118,7 @@ export class TsCache
 
 		if (noCache)
 		{
+			this.context.warn(yellow("Cleaning cache... If you want to use caching, please set `clean` option to false."));
```
agilgur5, line 121:
```suggestion
```

The correct location for the logging is in `index.ts`. You added that in the most recent commit, so can remove this

src/index.ts
```diff
@@ -82,11 +82,14 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 		cache.done();
 	}
 
+	const hasTransformersOrSourceMapCallback = options?.sourceMapCallback || options?.transformers?.length;
```
agilgur5, line 85:
```suggestion
	const hasFunctions = options?.sourceMapCallback || options?.transformers?.length;
```

Shorter and more durable name as more function options could be added like these two.
`hasFunctions` is also more meaningful in that sense, as to why these two options are special. Should probably add a comment above this stating that function scope can't be properly hashed, and linking to the `object-hash` issue

agilgur5 (changes requested):
I'm on my phone rn, so only left a few simpler comments for now.

With regard to where to use `extraCacheKeys`, it would be in the call to `object-hash` in `tscache`, i.e. in [this block](https://github.com/ezolenko/rollup-plugin-typescript2/blob/0b99f8ec62119d07df121beb441074b282d546db/src/tscache.ts#L94).

src/index.ts
```diff
@@ -149,6 +153,10 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 			service = tsModule.createLanguageService(servicesHost, documentRegistry);
 			servicesHost.setLanguageService(service);
 
+			if(!pluginOptions.clean && shouldEnableDefaultClean) {
+				context.warn("You have enabled transformers or sourceMapCallback, but have disabled the default clean option. You may need to use 'extraCacheKeys' option to enable it again.")
```
agilgur5, line 157:
```suggestion
				context.warn("You have enabled 'transformers' or 'sourceMapCallback'. To enable caching, you will have to configure the 'extraCacheKeys' option. See https://github.com/ezolenko/rollup-plugin-typescript2#plugin-options")
```

what the default is or isn't doesn't need to be specified. Just that caching is disabled. This is user-facing, so we want to make the message as simple and to-the-point as possible.

_Ideally_ we should also say _which_ of the two (or both) are enabled, as that makes for a better/more informative warning.

agilgur5 (changes requested):
a few other things from my laptop now

PR #424, opened 2022-09-18
fix: add compatibility checks w/ `semver`

### LINKED ISSUE #9: Detect incompatible Rollup versions
v0.36 doesn't have some entry points the plugin uses.

Should fail gracefully.

agilgur5:
## Summary

Add several compatibility checks using `semver`
- Follow-up to https://github.com/ezolenko/rollup-plugin-typescript2/pull/406#issuecomment-1250165068
- Fixes #9

## Details

A few commits here:
- **hotfix**: `this.load` was added in Rollup `2.60.0`; don't run on older versions
  - this is my bad, I didn't realize `this.load` came out [_much_ later](https://github.com/rollup/rollup/blob/master/CHANGELOG.md#2600) than `this.resolve`
  - since we're backward-compatible w/ Rollup `1.26.3`, add in a version check for the type-only fix
    - the type-only fix will error out in Rollup versions less than `2.60.0`, so instead of erroring out, just skip this functionality
  - TODO: should move back to using `semver` that was removed in 08d2f5bcad7cb9a9a0974f3707a27d29dd89da7c

- **deps**: add back `semver` and use for version checking
  - reverts the `semver` removal in 08d2f5bcad7cb9a9a0974f3707a27d29dd89da7c
    - add back the TS version range check
  - add a Rollup version range check with `$ROLLUP_VERSION_RANGE`, fixes #9
  - check `this.load` Rollup version with `semver` instead of string comparison
  - add `semver` to `external`s list
    - also remove `resolve` as it's no longer used as of #367
    - and re-organize the list so that Node built-ins are in one section while deps are in another
      - makes it clearer what we're marking as external

- **dx**: add a nice warning when `this.load` is not supported
  - so it doesn't just silently skip the type-only fix on Rollup versions <2.60.0
  
- **fix(test)**: add globals for strings that are replaced during build
  - `TS_VERSION_RANGE`, `ROLLUP_VERSION_RANGE`, `RPT2_VERSION` were previously only defined during builds, so would cause tests to fail
    - add these vars as (namespaced) globals in the Jest config so that they can be used in tests too
    - if they don't exist on `global` (or if `global` doesn't exist), then use the built strings instead
    - see in-line comments for more details
  - also reorder `re` placement to match the usage order in the code
  - and fix lint issues (`no-string-literal`)
  
## Review Notes

Put a few commits together b/c these are all very related (would be dependent on each other) and only make a few changes to the source code anyway.

I also built the `dist/` files to make sure that adding `semver` and `$ROLLUP_VERSION_RANGE` works. **NOTE** that this may need to be rebuilt prior to release if other PRs are merged in

Oddly enough, adding `semver` as an `external` removes it from the builds. When not an `external` though, it gets _added_ to the builds (additions in the `git diff`).... Not sure what's going on there exactly... Maybe it's only partially rolled up currently?
`package-lock.json` changes shows it was in the dependency tree already, but I only saw it in devDeps 🤔 Not sure why it was in the build.
Making it `external` makes it more logical at least (and allows users to override the dep etc per https://github.com/ezolenko/rollup-plugin-typescript2/pull/80#issuecomment-584451607)


agilgur5:
Ack, integration tests fail because the `$TS_VERSION_RANGE`, `$ROLLUP_VERSION_RANGE` etc are not substituted yet in the tests (they're only replaced in the built files)... will need to handle that somehow...

agilgur5:
Fixed the tests and lint issues now

Issue #437, opened 2023-02-15
`moduleResolution` is overridden 


zemse:
## Troubleshooting

1. Does `tsc` have the same output? If so, please explain why this is incorrect behavior
    

> `tsc` uses actual moduleResolution from my tsconfig so it works.

2. Does your Rollup plugin order match [this plugin's compatibility](https://github.com/ezolenko/rollup-plugin-typescript2#compatibility)? If not, please elaborate
    

> I'm using [tsdx](https://tsdx.io/customization#rollup), which uses this plugin internally, so not really sure about this.

3. Can you create a [minimal example](https://stackoverflow.com/help/minimal-reproducible-example) that reproduces this behavior? Preferably, use [this environment](https://stackblitz.com/edit/rpt2-repro) for your reproduction
    

> I can do that if needed. But this issue is very easily reproducable.

## What happens and why it is incorrect

I want to import something from a package that uses [package.json exports](https://www.typescriptlang.org/docs/handbook/esm-node.html). Here, ethers is a package, and address is defined as an export in it's [package.json](https://github.com/ethers-io/ethers.js/blob/9b508a46567babca45268774b89a63c90bdd0ef2/package.json#L46).

```ts
import type { AddressLike } from 'ethers/address';
```

The above compiles successfully with vanilla typescript compiler, when moduleResolution is set to node16. With rollup-plugin-typescript2, the moduleResolution is overridden and hence typescript compiler gives the exact error as if moduleResolution is set to `node`:

```
(typescript) Error: /my/project/src/file/path.ts(5,34): semantic error TS2307: Cannot 
find module 'ethers/address' or its corresponding type declarations.
```

The `moduleResolution` from user's tsconfig is overridden is set to legacy node resolution ([relevant code](https://github.com/ezolenko/rollup-plugin-typescript2/blob/f6db59613a66f58c48310aa8fa785951970b5d6d/src/get-options-overrides.ts#L19
)). I think this is incorrect because typescript wants moduleResolution to be `node16` or `nodenext` if a dev wants to use package json exports.

### Versions

```text
  System:
    OS: macOS 12.0.1
    CPU: (8) arm64 Apple M1
    Memory: 111.16 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.17.1 - ~/.nvm/versions/node/v16.17.1/bin/node
[...]

PR #447, opened 2023-05-20
updating module overrides

ezolenko:
## Summary

Fix for #437

## Details

- keeping node* `moduleResolution` option as is ("node", "node16", "nodenext")
- overriding "classic" and empty to Node10 (old behavior was overriding everything to Node10)
- overriding "bundler" to Node16


dist/parse-tsconfig.d.ts
```diff
@@ -1,7 +1,7 @@
-import { RollupContext } from "./context";
-import { IOptions } from "./ioptions";
-export declare function parseTsConfig(context: RollupContext, pluginOptions: IOptions): {
-    parsedTsConfig: import("typescript").ParsedCommandLine;
-    fileName: string | undefined;
-};
```
agilgur5, line 6: these look to be exactly the same, a little confused why it's showing as a diff? 🤔  
looks to be the same spacing too?

seems to have happened for several of the generated declarations

src/get-options-overrides.ts
```diff
@@ -16,13 +16,28 @@ export function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot }: IO
...
+		case undefined:
+			overrides.moduleResolution = tsModule.ModuleResolutionKind.Node10;
+			break;
+		case tsModule.ModuleResolutionKind.Bundler:
+		default:
+			overrides.moduleResolution = tsModule.ModuleResolutionKind.Node16;
+	}
+
 	if (preParsedTsconfig.options.module === undefined)
 		overrides.module = tsModule.ModuleKind.ES2015;
```
agilgur5, line 42: I believe this line will never get hit now / is dead code? test coverage should confirm that

agilgur5:
Replaced by #453

agilgur5:
Or rather, this got split into multiple PRs it looks like:
- #453 for `moduleResolution` changes
- #451 for `cts` + `mts`
- #450 for `module: 'ES2022'`

Much nicer and easier to read as separate PRs!

dist/parse-tsconfig.d.ts
agilgur5: self resolved as this PR was closed and this change doesn't exist in the other PRs

src/get-options-overrides.ts
agilgur5: oh I misread this, my bad. `module` vs. `moduleResolution` above

ezolenko:
Yeah, ignore that PR, I started too many changes at once and updated deps too far and that required yet more changes and so on (we'd still have to figure out ts-jest problems eventually I guess). All interesting changes are pulled out into those 3 PRs

agilgur5:
Ah right, `ts-jest` is having issues because `jest`'s ESM support is still [experimental](https://jestjs.io/docs/ecmascript-modules), meaning some CJS is still needed. `ts-jest` now has [some support](https://kulshekhar.github.io/ts-jest/docs/guides/esm-support/) too now though, so might be able to get it to work that way.

**EDIT**: welp, that gives even more problems because some of the test code does not work in ESM mode (e.g. `__dirname`, `require` for `tslib`, and imports of CJS packages)

Issue #448, opened 2023-05-20
how can we set the output `index.d.ts` file to be `index.d.cts` for the `cjs` format instead?


mesqueeb:
### currently

With this plugin I generate `index.cjs` via rollup and this plugin generates the `index.d.ts` next to it.

### desired

how can I set the generated `index.d.ts` file to be called `index.d.cts` instead?

### Versions

```text
  System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M2
    Memory: 309.73 MB / 24.00 GB
    Shell: 3.6.0 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.20.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.4 - /usr/local/bin/npm
  npmPackages:
    rollup: ^3.22.0 => 3.22.0
    rollup-plugin-typescript2: ^0.34.1 => 0.34.1
    typescript: ^5.0.4 => 5.0.4
```

<details>
  <summary><h4><code>rollup.config.js</code></h4>: </summary>

```js
import typescript from 'rollup-plugin-typescript2'
import pkg from '../package.json' assert { type: 'json' }

export default {
  input: 'src/index.ts',
  output: [
    { file: pkg.exports['.'].require.default, format: 'cjs' },
    { file: pkg.exports['.'].import.default, format: 'esm' },
  ],
  external: Object.keys(pkg.dependencies || []),
  plugins: [typescript({ tsconfigOverride: { exclude: ['test/**/*'] } })],
}

```

</details>

<details>
  <summary><h4><code>tsconfig.json</code></h4>: </summary>

```json5
{
  "compilerOptions": {
    "allowJs": false,
    "baseUrl": ".",
    "declaration": true,
    "downlevelIteration": true,
    "esModuleInterop": true,
    "isolatedModules": true,
    "lib": ["esnext", "DOM", "DOM.Iterable"],
    "module": "esnext",
    "moduleResolution": "node",
    "noImplicitReturns": true,
    "noUnusedLocals": true,
    "skipLibCheck": true,
    "sourceMap": true,
    "strict": true,
    "target": "es2019",
    "useDefineForClassFields": true
  },
  "include": ["src/**/*", "test/**/*"]
}

```

</details>

<details>
  <summary><h4><code>package.json</code></h4>: </summary>

```json
  "sideEffects": false,
  "type": "module",
[...]

Issue #446, opened 2023-05-19
`Unexpected token` when importing `*.d.cts` or `*.d.mts` files


phun-ky:
## Troubleshooting

1. Does `tsc` have the same output? If so, please explain why this is incorrect behavior

    N/A

1. Does your Rollup plugin order match [this plugin's compatibility](https://github.com/ezolenko/rollup-plugin-typescript2#compatibility)? If not, please elaborate

    yes

1. Can you create a [minimal example](https://stackoverflow.com/help/minimal-reproducible-example) that reproduces this behavior? Preferably, use [this environment](https://stackblitz.com/edit/rpt2-repro) for your reproduction
    

    https://stackblitz.com/edit/rpt2-repro-3j5vny?file=src%2Findex.ts,package.json,src%2Flib%2Findex.ts,README.md,tsconfig.json

## What happens and why it is incorrect

Error is thrown, should have support for `*.cts`/`*.mts` files

```
[!] (plugin rpt2) RollupError: Unexpected token (Note that you need plugins to import files that are not JavaScript)
../../node_modules/helmet/index.d.cts (3:5)
1: import {IncomingMessage, ServerResponse} from "http"
2: 
3: type ContentSecurityPolicyDirectiveValueFunction = (req: IncomingMessage, res: ServerResponse) => string
        ^
4: type ContentSecurityPolicyDirectiveValue = string | ContentSecurityPolicyDirectiveValueFunction
5: interface ContentSecurityPolicyOptions {

    at error (/home/alexander/Workspace/company/internal/devplatform-server-api/node_modules/rollup/dist/shared/rollup.js:279:30)
    at Module.error (/home/alexander/Workspace/company/internal/devplatform-server-api/node_modules/rollup/dist/shared/rollup.js:13947:16)
    at Module.tryParse (/home/alexander/Workspace/company/internal/devplatform-server-api/node_modules/rollup/dist/shared/rollup.js:14658:25)
    at Module.setSource (/home/alexander/Workspace/company/internal/devplatform-server-api/node_modules/rollup/dist/shared/rollup.js:14260:39)
    at ModuleLoader.addModuleSource (/home/alexander/Workspace/company/internal/devplatform-server-api/node_modules/rollup/dist/shared/rollup.js:24225:20)
```
```
[!] (plugin rpt2) RollupError: Unexpected token (Note that you need plugins to import files that are not JavaScript)
../../node_modules/helmet/index.d.mts (3:5)
[...]

PR #451, opened 2023-06-20
cts/mts support

ezolenko:
## Summary

Fix for #446

## Details

Adding cts/mts to the known filetypes


src/index.ts
```diff
@@ -274,7 +274,7 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 			// Rollup can't see these otherwise, because they are "emit-less" and produce no JS
 			if (result.references && supportsThisLoad) {
 				for (const ref of result.references) {
-					if (ref.endsWith(".d.ts"))
```
agilgur5, line 277: So I believe I _intentionally_ wrote this as hard-coded, to match [this line](https://github.com/ezolenko/rollup-plugin-typescript2/blob/68017aeb9eb81f9473bfe037ece52a837c1986b5/src/index.ts#L210) in `resolveId`. `filter` is still called after that line. 

I believe the intent of that is that rpt2 should _never_ resolve declarations, even if a user overrides the `exclude`.

In the branch I mentioned in https://github.com/ezolenko/rollup-plugin-typescript2/issues/446#issuecomment-1625859100, similar to `resolveId`, I do _both_ a declaration check _and_ then a `filter` check. I then had a [follow-up commit](https://github.com/agilgur5/rollup-plugin-typescript2/commit/15bd6ec3bbf448aac7c02d69c3a360a256724fa3) to refactor those checks into a single function to always be equivalent.

Let me make a PR for that. I'll expand that function to also check all current declaration extensions (i.e. `.d.ts` _as well as_ `.d.cts` and `.d.mts`)

agilgur5: Added #456 to fix that

src/index.ts
```diff
@@ -95,8 +95,8 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 			verbosity: VerbosityLevel.Warning,
 			clean: false,
 			cacheRoot: findCacheDir({ name: "rollup-plugin-typescript2" }),
-			include: ["*.ts+(|x)", "**/*.ts+(|x)"],
```
agilgur5, line 98: I think we need to modify [the TS emit output conversion](https://github.com/ezolenko/rollup-plugin-typescript2/blob/68017aeb9eb81f9473bfe037ece52a837c1986b5/src/tscache.ts#L38) to also handle `.d.cts`, `.d.mts` and their respective mappings. I _assume_ that TS would take a CTS file and output `.cjs`, `d.cts`, and `d.cts.map` files, which we would not handle correctly right now.

We should also definitely have integration tests for CTS / MTS, especially to validate that assumption above^

ezolenko: Yeah it would fix #448 too

agilgur5: Oh I didn't even think about that, but it would indeed fix #448. There is a slight problem there with declarations though; Rollup cannot convert them between `.d.cts` and `.d.mts`, so they are limited to the user's original module format. I'll detail that more in the issue itself.

agilgur5: For the emit output conversion, I looked up the source code and indeed `OutputFile` now has some new extensions:
- [`.mjs` + `.cjs`](https://github.com/microsoft/TypeScript/blob/d8585688dd1bc8d82b7b5daab9af83ae1e3de197/src/compiler/emitter.ts#L576)
- [`.d.cts`, `.d.mts`](https://github.com/microsoft/TypeScript/blob/d8585688dd1bc8d82b7b5daab9af83ae1e3de197/src/compiler/utilities.ts#L6229)
- `.map` files of those are just [appended](https://github.com/microsoft/TypeScript/blob/d8585688dd1bc8d82b7b5daab9af83ae1e3de197/src/compiler/emitter.ts#L546) to the existing extension from above

I imagine that with `module` set to output ESM (for Rollup), we'd only see the `.mjs` variants, but testing can confirm that

PR #452, opened 2023-06-20
disabling cache in watch mode

### LINKED ISSUE #443: cache use hash cost last of time ，cause it too slow when rebuild in watch mode
## Troubleshooting

1. Does `tsc` have the same output? If so, please explain why this is incorrect behavior
    

1. Does your Rollup plugin order match [this plugin's compatibility](https://github.com/ezolenko/rollup-plugin-typescript2#compatibility)? If not, please elaborate
    

1. Can you create a [minimal example](https://stackoverflow.com/help/minimal-reproducible-example) that reproduces this behavior? Preferably, use [this environment](https://stackblitz.com/edit/rpt2-repro) for your reproduction
    

### Versions

```text
System:
    OS: Linux 4.19 Debian GNU/Linux 10 (buster) 10 (buster)
    CPU: (8) x64 Intel(R) Xeon(R) Platinum 8260 CPU @ 2.40GHz
    Memory: 10.49 GB / 15.41 GB
    Container: Yes
    Shell: 5.7.1 - /usr/bin/zsh
 Binaries:
    Node: 16.19.0 - /run/user/1001/fnm_multishells/846809_1680868810665/bin/node
    Yarn: 1.22.19 - /run/user/1001/fnm_multishells/846809_1680868810665/bin/yarn
    npm: 8.19.3 - /run/user/1001/fnm_multishells/846809_1680868810665/bin/npm
  npmPackages:
    typescript: ^5.0.2 => 5.0.2 
```

<details>
  <summary><h4><code>rollup.config.js</code></h4>: </summary>

```js
import path from 'path';
import { defineConfig } from 'rollup';

import { genPlugin, genTypeDefinePlugin } from './plugins';
import { resolveFile, getPackageJson } from './utils';

const pkg = getPackageJson();
const libName = pkg.name;

const devPlugins = genPlugin({
  isDev: true,
  pluginTypescriptOptions: {
    sourceMap: true,
  },
});

/** @type {import('rollup').OutputOptions[]} */
const output = [
  // {
  //   file: resolveFile(`dist/${libName}.cjs.js`),
  //   format: 'cjs',
  // },
  {
    file: resolveFile(`dist/${libName}.js`),
    format: 'es',
  },
];

// build config for development
export default defineConfig([
  {
    watch: {
      include: ['src/**'],
    },
    input: resolveFile('src/index.ts'),
    output,
    plugins: devPlugins,
    external: [...Object.keys(pkg.dependencies), 'protobufjs/minimal'],
  },
]);

```

</details>

<details>
  <summary><h4><code>tsconfig.json</code></h4>: </summary>

```json5
{
  "compilerOptions": {
    "moduleResolution": "node",
    "module": "ESNext",
    "target": "ES2020",
    "lib": ["dom", "ESNext"],
    "sourceMap": true,
    "declaration": true,
    "declarationMap": true,
    "declarationDir": "./type",
    "outDir": "./lib",
    "strict": true,
    "noImplicitAny": true,
    "strictPropertyInitialization": false,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "baseUrl": ".",
    "paths": {
      "@/*": ["src/*"]
    },
    "plugins": [
      {
        "transform": "typescript-transform-paths",
        "useRootDirs": true,
        "afterDeclarations": true
      }
    ]
  },
  "include": ["src/**/*.ts", "src/**/*.d.ts"]
}

```

</details>

<details>
  <summary><h4><code>package.json</code></h4>: </summary>

```json
 "rollup-plugin-typescript2": "^0.34.1",
```

</details>

<details>
  <summary><h4>plugin output with verbosity 3</h4>: </summary>

```text
```
</details>

## detail

```text
// when rebuild  rollup-plugin-typescript2 will compute hash key for cache
// but rollupConfig is so big object and cost so many time in compute hash
this.cacheDir = `${this.cacheRoot}/${this.cachePrefix}${objHash({
          version: this.cacheVersion,
          rootFilenames,
          options: this.options,
          rollupConfig: this.rollupConfig,
          tsVersion: tsModule.version,
      }, this.hashOptions)}`;
```

image

image

in profile，we can see hash compute  almost   cost all time, about 20s

### LINKED ISSUE #444: Watch process not updating file
### Rollup Version

3.20.2

### Operating System (or Browser)

Windows 10 latest

### Node Version (if applicable)

18.13.0

### Link To Reproduction

https://stackblitz.com/edit/rollup-repro-jz3iti?file=package.json
OR
https://github.com/nfxpnk/rollup-test

### Expected Behaviour

1. Run watch process: `npm run watch`
2. Edit file `test/flying-saucer/base/global/index.ts`
3. Watch process updates
4. See builded file - `dist-test/start.js`
5. New changes made in `test/flying-saucer/base/global/index.ts` present in builded file `dist-test/start.js`

### Actual Behaviour

1. Run watch process: `npm run watch`
2. Edit file `test/flying-saucer/base/global/index.ts`
3. Watch process updates
4. See builded file - `dist-test/start.js`
5. No changes in `dist-test/start.js`

If you will run just regular build with `npm run build` it works as expected.

Clean JS - not reproduceable - https://stackblitz.com/edit/rollup-repro-eghflu?file=rollup.config.js

ezolenko:
## Summary

Fixes #443, should incidentally fix #444

## Details

Disabling cache in watch mode, speeds it up considerably, which was the whole point of having a cache in the first place.

This is slightly different from simply using `clean: true` for watch mode -- `clean: true` also deletes all old caches. `clean: false` in watch mode would bypass cache without touching old caches.

### DIFF
--- a/src/tscache.ts
+++ b/src/tscache.ts
@@ -78,16 +78,16 @@ export class TsCache
 	private syntacticDiagnosticsCache!: ICache<IDiagnostics[]>;
 	private hashOptions = { algorithm: "sha1", ignoreUnknown: false };
 
-	constructor(private noCache: boolean, hashIgnoreUnknown: boolean, private host: tsTypes.LanguageServiceHost, private cacheRoot: string, private options: tsTypes.CompilerOptions, private rollupConfig: any, rootFilenames: string[], private context: RollupContext)
+	constructor(private noCache: boolean, runClean: boolean, hashIgnoreUnknown: boolean, private host: tsTypes.LanguageServiceHost, private cacheRoot: string, private options: tsTypes.CompilerOptions, private rollupConfig: any, rootFilenames: string[], private context: RollupContext)
 	{
 		this.dependencyTree = new Graph({ directed: true });
 		this.dependencyTree.setDefaultNodeLabel((_node: string) => ({ dirty: false }));
 
-		if (noCache)
-		{
+		if (runClean)
 			this.clean();
+
+		if (noCache)
 			return;
-		}
 
 		this.hashOptions.ignoreUnknown = hashIgnoreUnknown;
 		this.cacheDir = `${this.cacheRoot}/${this.cachePrefix}${objHash(
--- a/src/index.ts
+++ b/src/index.ts
@@ -166,7 +166,10 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
 			service = tsModule.createLanguageService(servicesHost, documentRegistry);
 			servicesHost.setLanguageService(service);
 
-			cache = new TsCache(pluginOptions.clean, pluginOptions.objectHashIgnoreUnknownHack, servicesHost, pluginOptions.cacheRoot, parsedConfig.options, rollupOptions, parsedConfig.fileNames, context);
+			const runClean = pluginOptions.clean;
+			const noCache = pluginOptions.clean || watchMode;
+
+			cache = new TsCache(noCache, runClean, pluginOptions.objectHashIgnoreUnknownHack, servicesHost, pluginOptions.cacheRoot, parsedConfig.options, rollupOptions, parsedConfig.fileNames, context);
 
 			// reset transformedFiles Set on each watch cycle
 			transformedFiles = new Set<string>();


src/tscache.ts
```diff
@@ -78,16 +78,16 @@ export class TsCache
 	private syntacticDiagnosticsCache!: ICache<IDiagnostics[]>;
 	private hashOptions = { algorithm: "sha1", ignoreUnknown: false };
 
-	constructor(private noCache: boolean, hashIgnoreUnknown: boolean, private host: tsTypes.LanguageServiceHost, private cacheRoot: string, private options: tsTypes.CompilerOptions, private rollupConfig: any, rootFilenames: string[], private context: RollupContext)
+	constructor(private noCache: boolean, runClean: boolean, hashIgnoreUnknown: boolean, private host: tsTypes.LanguageServiceHost, private cacheRoot: string, private options: tsTypes.CompilerOptions, private rollupConfig: any, rootFilenames: string[], private context: RollupContext)
```
agilgur5, line 81: `runClean` should also be marked `private`

src/tscache.ts
```diff
@@ -78,16 +78,16 @@ export class TsCache
...
 	{
 		this.dependencyTree = new Graph({ directed: true });
 		this.dependencyTree.setDefaultNodeLabel((_node: string) => ({ dirty: false }));
 
-		if (noCache)
-		{
+		if (runClean)
```
agilgur5, line 86: Does skipping `this.clean()` have a significant performance improvement?
I imagine that after the first clean, the performance hit would be negligible

src/tscache.ts
ezolenko: It's not used outside of constructor itself, so no need to make it a member.

src/tscache.ts
ezolenko: Yeah, this is more to avoid surprising behavior -- cleaning old caches and disabling caching altogether are different actions, but they were joined together before

src/tscache.ts
agilgur5: Oh my bad, I forgot how this syntax worked and thought it was `public` by default (it is not, [docs](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties)). I rarely use classes, sorry about that 😅

src/tscache.ts
agilgur5: > cleaning old caches and disabling caching altogether are different actions

Yea `clean` is a bit of an edge-case as it's a side-effect. Ideally one would `clean` in a totally separate command/export, but there's no conventional way of doing that in Rollup. So while a bit of a head-scratcher initially, it did eventually make sense to me. When I was optimizing the cache, I thought about changing this for a bit but ended up leaving it as is.

> to avoid surprising behavior

To an experienced user, I thought it might be surprising that something runs without caching and _doesn't_ clean. Since they have been joined at the hip for so long.

But this turns off caching _implicitly_ with no config, so I suppose we can ignore that scenario.
If `watch` removed caches when it previously didn't, that would probably be _more_ surprising.

Had to write the logic out there a bit, but that makes sense to me now. Removing a head-scratcher (partially) required some head-scratching 😅

<hr>

We could potentially expose this as an option to completely eliminate head-scratching so that `cleanCache` and `noCache` are entirely separate config options 🤔 It would technically be a breaking change though to fully separate those.

When I was reviewing this, I remembered there was some reason I never exposed it myself. I think it was potential bugginess of the dependency tree with no cache, but I think I might have fixed all those code paths in #362 et al... Testing my memory a bit here 😅

Issue #434, opened 2023-01-26
ESM build error on directory imports after upgrade to TS 5.0.0@beta


bluebill1049:
## Troubleshooting

1. Does `tsc` have the same output? If so, please explain why this is incorrect behavior
    
    It's working correctly with `tsc`

1. Does your Rollup plugin order match? If not, please elaborate
    
    Rollup config hasn't been changed for years and looks like the order is not an issue:
https://github.com/react-hook-form/react-hook-form/blob/master/scripts/rollup/createRollupConfig.js#L14-L44

1. Can you create a [minimal example](https://stackoverflow.com/help/minimal-reproducible-example) that reproduces this behavior? Preferably, use [this environment](https://stackblitz.com/edit/rpt2-repro) for your reproduction
    
    
    Here is the PR: https://github.com/react-hook-form/react-hook-form/pull/9834

## What happens and why it is incorrect

Related to `export *`

```
> react-hook-form@7.42.1 build:esm /home/runner/work/react-hook-form/react-hook-form
> rollup -c ./scripts/rollup/rollup.esm.config.js


src/index.ts → dist/index.esm.mjs...
[!] (plugin rpt2) Error: src/index.ts:2:15 - error TS2792: Cannot find module './logic'. Did you mean to set the 'moduleResolution' option to 'nodenext', or to add aliases to the 'paths' option?

2 export * from './logic';
                ~~~~~~~~~
src/index.ts:3:15 - error TS2792: Cannot find module './types'. Did you mean to set the 'moduleResolution' option to 'nodenext', or to add aliases to the 'paths' option?

3 export * from './types';
                ~~~~~~~~~
src/index.ts:10:15 - error TS2792: Cannot find module './utils'. Did you mean to set the 'moduleResolution' option to 'nodenext', or to add aliases to the 'paths' option?

10 export * from './utils';
                 ~~~~~~~~~

src/index.ts

    at error (/home/runner/work/react-hook-form/react-hook-form/node_modules/.pnpm/rollup@2.79.1/node_modules/rollup/dist/shared/rollup.js:19[8](https://github.com/react-hook-form/react-hook-form/actions/runs/4019858069/jobs/6907177794#step:6:9):30)
[...]

PR #453, opened 2023-06-20
support newer `moduleResolution` kinds, update build to TS 5.x

ezolenko:
## Summary

Fix for #437

## Details

- keeping node* moduleResolution option as is ("node", "node16", "nodenext")
- overriding "classic" and empty to Node10 (old behavior was overriding everything to Node10)
- overriding "bundler" to Node10


src/get-options-overrides.ts
```diff
@@ -16,13 +16,26 @@ export function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot }: IO
...
+	{
+		case tsModule.ModuleResolutionKind.Node10:
+		case tsModule.ModuleResolutionKind.Node16:
+		case tsModule.ModuleResolutionKind.NodeNext:
+			overrides.moduleResolution = preParsedTsconfig.options.moduleResolution;
+			break;
+		case tsModule.ModuleResolutionKind.Classic:
+		case tsModule.ModuleResolutionKind.Bundler:
+		default:
+			overrides.moduleResolution = tsModule.ModuleResolutionKind.Node10;
```
agilgur5, line 36: I'm gonna need to review this in a lot more detail (need to catch up on TS changes since 5.x hit GA and refresh myself on this codebase), but I mentioned `mjs` support and `bundler` in https://github.com/ezolenko/rollup-plugin-typescript2/issues/434#issuecomment-1407796721.

[`bundler`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#moduleresolution-bundler) actually seemed the most appropriate at the time (since Rollup is a bundler) for `Node16`+, particularly due to the file extension issues. That's actually why I didn't implement support for `mjs` / `cjs` earlier because the file extension requirement could break many things (that and no user requested it)

ezolenko: Yeah, this basically lowest effort to let users use moduleResolution without figuring out exactly what the differences are.

agilgur5: So I think `bundler` should not be overwritten either. With that change and README modifications about this, I think we could approve and merge this in.

Without tests though, we won't be able to confirm some of the behavior of `node16` / `nodeNext` -- it is possible we may need to override those two to `bundler` as well

src/get-options-overrides.ts
```diff
@@ -16,13 +16,26 @@ export function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot }: IO
 		noEmitOnError: false,
 		inlineSourceMap: false,
 		outDir: normalize(`${cacheRoot}/placeholder`), // need an outdir that is different from source or tsconfig parsing trips up. https://github.com/Microsoft/TypeScript/issues/24715
-		moduleResolution: tsModule.ModuleResolutionKind.NodeJs,
 		allowNonTsExtensions: true,
+		moduleResolution: tsModule.ModuleResolutionKind.Node10,
```
agilgur5, line 20: mmm, I think it would be more straightforward logic to have _no_ override by default. we would only add an override in the case of `classic`

src/get-options-overrides.ts
```diff
@@ -16,13 +16,26 @@ export function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot }: IO
...
+		case tsModule.ModuleResolutionKind.Node16:
+		case tsModule.ModuleResolutionKind.NodeNext:
+			overrides.moduleResolution = preParsedTsconfig.options.moduleResolution;
+			break;
+		case tsModule.ModuleResolutionKind.Classic:
+		case tsModule.ModuleResolutionKind.Bundler:
+		default:
+			overrides.moduleResolution = tsModule.ModuleResolutionKind.Node10;
+	}
+
```
agilgur5, line 38: with no override above, could simplify this to:
```suggestion
	if (preParsedTsconfig.options.moduleResolution === tsModule.ModuleResolutionKind.Classic)
	  overrides.moduleResolution = tsModule.ModuleResolutionKind.Node10;
```

src/get-options-overrides.ts
ezolenko: Yeah, if we pass bundler through the only problem could be new types in the future, but we can deal with it in readme.

agilgur5: Or we could add a check similar to the `module` one.
But I'm not particularly worried about forward compat, future changes may require future changes after all.

This line still needs to be removed

README.md
```diff
@@ -48,7 +48,7 @@ This also allows for passing in different `tsconfig` files depending on your bui
 * `inlineSourceMap`: false (see [#71](https://github.com/ezolenko/rollup-plugin-typescript2/issues/71))
 * `outDir`: `./placeholder` in cache root (see [#83](https://github.com/ezolenko/rollup-plugin-typescript2/issues/83) and [Microsoft/TypeScript#24715](https://github.com/Microsoft/TypeScript/issues/24715))
 * `declarationDir`: Rollup's `output.file` or `output.dir` (*unless `useTsconfigDeclarationDir` is true in the plugin options*)
-* `moduleResolution`: `node` (*`classic` is [deprecated](https://www.typescriptlang.org/docs/handbook/module-resolution.html). It also breaks this plugin, see [#12](https://github.com/ezolenko/rollup-plugin-typescript2/issues/12) and [#14](https://github.com/ezolenko/rollup-plugin-typescript2/issues/14)*)
+* `moduleResolution`: *`classic` is [deprecated](https://www.typescriptlang.org/docs/handbook/module-resolution.html) and changed to `node`. It also breaks this plugin, see [#12](https://github.com/ezolenko/rollup-plugin-typescript2/issues/12) and [#14](https://github.com/ezolenko/rollup-plugin-typescript2/issues/14)*). All other values are passed through. Rollup might have problems with some of them, so if in doubt, use `node`.
```
agilgur5, line 51: I think it makes more sense to move this down to "some compiler options have more than one compatible value". Then we can enumerate the current options there, and at the end specify why `classic` is not one of the compatible ones.

README.md
```diff
@@ -48,13 +48,17 @@ This also allows for passing in different `tsconfig` files depending on your bui
...
 * `allowNonTsExtensions`: true to let other plugins on the chain generate typescript; update plugin's `include` filter to pick them up (see [#111](https://github.com/ezolenko/rollup-plugin-typescript2/issues/111))
 
 ### Some compiler options have more than one compatible value
 
 * `module`: defaults to `ES2015`. Other valid values are `ES2020`, `ES2022` and `ESNext` (required for dynamic imports, see [#54](https://github.com/ezolenko/rollup-plugin-typescript2/issues/54)).
 
+* `moduleResolution`: defaults to `node10` (same as `node`), but value from tsconfig is used if specified. Other valid (but mostly untested) values are `node16`, `nodenext` and `bundler`. If in doubt, use `node10`.
+
+	`classic` is [deprecated](https://www.typescriptlang.org/docs/handbook/module-resolution.html) and changed to `node10`. It also breaks this plugin, see [#12](https://github.com/ezolenko/rollup-plugin-typescript2/issues/12) and [#14](https://github.com/ezolenko/rollup-plugin-typescript2/issues/14).
+
```
agilgur5, line 60: I think this looks a bit nicer with a sub-bullet, which I've used in the release notes before (stylistically, as an "additional note").

```suggestion
	* `classic` is [deprecated](https://www.typescriptlang.org/docs/handbook/module-resolution.html) and changed to `node10`. It also breaks this plugin, see [#12](https://github.com/ezolenko/rollup-plugin-typescript2/issues/12) and [#14](https://github.com/ezolenko/rollup-plugin-typescript2/issues/14).
```

dist/rollup-plugin-typescript2.cjs.js
```diff
@@ -27834,11 +27834,22 @@ function getOptionsOverrides({ useTsconfigDeclarationDir, cacheRoot }, preParsed
         noEmitOnError: false,
         inlineSourceMap: false,
         outDir: pluginutils.normalizePath(`${cacheRoot}/placeholder`),
-        moduleResolution: tsModule.ModuleResolutionKind.NodeJs,
         allowNonTsExtensions: true,
+        moduleResolution: tsModule.ModuleResolutionKind.Node10,
     };
     if (!preParsedTsconfig)
         return overrides;
+    switch (preParsedTsconfig.options.moduleResolution) {
```
agilgur5, line 27842: needs a rebuild now

agilgur5 (commented):
mostly LGTM now, just left a tiny comment style comment. we'll ofc iterate on the docs as we test more and get feedback

tcschiller:
Hey team! Thanks for this. Wondering when it might get released / published to npm?

ezolenko:
@tcschiller could you test to see if this PR works for your use-case?

tcschiller:
> @tcschiller could you test to see if this PR works for your use-case?

Pulled down latest from the `master` branch and ran the following:

`npm ci`
`npm run build`
`npm run build-self`
`npm run build-self`
`npm pack`

Then I tried my own repo, first, _without_ that tarball. Built my library and successfully reproduced the problem at hand - it can't resolve imports with relative paths (e.g., `import { myExport } from @scope/package-name/submodule`).

Then I ran in my own repo:

`npm install ../rollup-plugin-typescript2/rollup-plugin-typescript2-0.35.1.tgz`

And immediately retried the build which, up to this point, was failing.

Could no longer reproduce the problem; it seems the MR discussed here fixes everything (because my TypeScript config has the `bundler` value for `moduleResolution`.

Thanks for the hard work, @ezolenko . Hope the above proves helpful!

agilgur5:
Thanks for testing this out @tcschiller!

For reference, you can reference a GH repo [directly in NPM](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#github-urls) (and any `git` repo as well) and use a SHA. This commit in particular made a change to the `dist` folder in the repo, so it already has the compiled version. Very useful for exactly this kind of scenario 🙂

agilgur5:
I was also curious about `node16` / `nodeNext` [above](https://github.com/ezolenko/rollup-plugin-typescript2/pull/453#discussion_r1260397170), specifically whether we should force `bundler` on those b/c of the file extension requirements.

I approved a few months ago since this is still better than the previous state, but that was the one remaining question on this PR

ezolenko:
Well, shouldn't break anything. I'll release this sometime tomorrow.

ezolenko:
In 0.36.0 now

PetarKirov:
@ezolenko with 0.36.0 I'm getting the following error:

```
[!] (plugin rpt2) Error: Incompatible tsconfig option. Module resolves to 'Node16'. This is incompatible with Rollup, please use 'module: "ES2015"', 'module: "ES2020"', 'module: "ES2022"', or 'module: "ESNext"'.
```

With the following tsconfig settings:

```json
{
  "compilerOptions": {
    "module": "Node16",
    "moduleResolution": "Node16",
    // ...
  }
}
```

And the following Rollup plugin config settings:

```js
  const fileExtensions = ['.mjs', '.cjs', '.js', '.ts', '.json'];

  const plugins = [
    json(),
    commonjs(),

    peerDepsExternal({
      packageJsonPath: path.join(dir, 'package.json'),
    }),
    nodeResolve({
      extensions: fileExtensions,
      preferBuiltins: true,
      exportConditions: ['node'],
      mainFields: ['module', 'main'], // This instructs Rollup to prioritize ESM over CJS when resolving modules
      modulesOnly: true,
    }),
    typescript({
      check: true,
      tsconfig: path.join(dir, 'tsconfig.lib.json'),
    }),
    del({ targets: `${dir}/dist/*` }),
  ];
  ```

I'm using the package following versions:
* `rollup`: `3.29.4`
* `@rollup/plugin-node-resolve`: `15.2.1`
* `rollup-plugin-typescript2`: `0.36.0`

As far as I can see, this comes from these lines:

https://github.com/ezolenko/rollup-plugin-typescript2/blob/0.36.0/src/parse-tsconfig.ts#L46-L48

agilgur5:
@PetarKirov this PR only changed the `moduleResolution` config, not the `module` config.

There are potentially additional changes needed for `module`, but for the most part, we'd likely just be overriding it to `ESNext` or something. As Rollup requires ESM throughout, CJS output would not be compatible (which TS can output CJS for [`module` `node16` and `nodenext`](https://www.typescriptlang.org/tsconfig#node16nodenext-nightly-builds))

raphael-theriault-swi:
This doesn't work with either `node16` or `nodenext` since they require the `module` field to match

PR #457, opened 2023-07-09
github: mark `dist/*` as generated code

agilgur5:
## Summary

Mark `dist/*` as generated code for GitHub
- per https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github and https://github.com/github-linguist/linguist/blob/master/docs/overrides.md#generated-code

## Details

- should help a good bit with making diffs more condensed by default
  - can still uncollapse generated code, just that default will be collapsed

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 * -text
+dist/* linguist-generated


ezolenko:
Oh, nice

Issue #435, opened 2023-02-04
Support for `module: 'ES2022'`


neilenns:
## Troubleshooting

1. Does `tsc` have the same output? If so, please explain why this is incorrect behavior

No.

1. Does your Rollup plugin order match [this plugin's compatibility](https://github.com/ezolenko/rollup-plugin-typescript2#compatibility)?

n/a

1. Can you create a [minimal example](https://stackoverflow.com/help/minimal-reproducible-example) that reproduces this behavior? Preferably, use [this environment](https://stackblitz.com/edit/rpt2-repro) for your reproduction

Yes. [Minimal repo environment](https://stackblitz.com/edit/rpt2-repro-ey59dp?file=tsconfig.json)

## What happens and why it is incorrect

When building using the node16-strictest-esm standard tsconfig the following error is generated:

`[!] (plugin rpt2) RollupError: Incompatible tsconfig option. Module resolves to 'ES2022'. This is incompatible with Rollup, please use 'module: "ES2015"', 'module: "ES2020"', or 'module: "ESNext"'.`

## Environment
n/a, reproduces in the minimal environment

### Versions
```text
  System:
    OS: Windows 10 10.0.22621
    CPU: (12) x64 AMD Ryzen 5 5600X 6-Core Processor
    Memory: 16.43 GB / 31.91 GB
  Binaries:
    Node: 16.19.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.19.3 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    rollup: ^3.13.0 => 3.13.0
    rollup-plugin-typescript2: ^0.34.1 => 0.34.1
    typescript: ^4.9.5 => 4.9.5
```

<details>
  <summary><h4><code>rollup.config.js</code></h4>: </summary>

```js
import typescript from 'rollup-plugin-typescript2';
import resolve from '@rollup/plugin-node-resolve';
import css from 'rollup-plugin-import-css';
import copy from 'rollup-plugin-copy';
import commonjs from '@rollup/plugin-commonjs';

export default {
  input: 'MyInstrument.tsx',
  output: {
    dir: 'build',
    format: 'es'
  },
  plugins: [
    css({ output: 'MyInstrument.css' }), 
    resolve(), 
    commonjs(),
    typescript(),
    copy({
      targets: [
        { src: '*.html', dest: 'build'}
      ]
    })
  ],

[...]

PR #458, opened 2023-07-14
deps: bump Rollup peerDep to 3.7.5+, remove tech debt

agilgur5:
## Summary

Bumping the Rollup peerDep and removing legacy tech debt workarounds
- As discussed in https://github.com/ezolenko/rollup-plugin-typescript2/issues/435#issuecomment-1629852786

### Rollup upgrade

- `3.7.5` includes https://github.com/rollup/rollup/pull/4749
  - which fixes the duplicate error logging upstream and allows us to remove the `buildEnd` workaround
- `2.60.0` includes `this.load`, so can remove the `satisfies` check
- `2.14.0` includes `this.meta.watchMode`, so can remove the env check

### Other tech debt removals

Figured I'd remove some other deprecated code / tech debt while I'm at it
- remove deprecated `rollupCommonJSResolveHack`
  - it hasn't done anything since #251 in late 2020 (~2.5 years ago)
  - and has been formally deprecated since #367 over a year ago
- remove `objectHashIgnoreUnknownHack` warning
  - hasn't been needed for async functions since #203 in early 2020 (~3.5 years ago) - so I think that's a long enough window to now remove the warning
  - also add a link in the docs to `object-hash`
    - noticed there wasn't one, despite all the links I added to the docs!


src/index.ts
```diff
@@ -312,19 +301,8 @@ const typescript: PluginImpl<RPT2Options> = (options) =>
...
-				if (stackOnly)
-					this.error({ ...err, message: err.message, stack: stackOnly });
-				else
-					this.error(err);
-			}
-
-			if (!pluginOptions.check)
-				return buildDone();
+			if (err || !pluginOptions.check)
+			  return buildDone();
```
agilgur5, line 305: the logic here is a bit confusing -- I had to re-read my own code a few times.
Basically, I made sure the diff was _functionally_ equivalent to the previous one, minus the error rewriting.

`this.error` causes an exit and `return` will similarly end the function.

If there's an error, we just want to call `cache.done()` (as `buildDone()` does) and otherwise let the error proceed.
If we're not doing type-checking, we can return early as the code below is only relevant when type-checking

agilgur5:
Left this as a Draft PR for now as it is breaking, and I think we should get in a patch release and stabilize MTS/CTS support beforehand.

#453 would be ideal to get in beforehand as well, although that is technically a breaking change as well

agilgur5:
ooo tests caught that our devDeps need updating too. updating Rollup to 3+ requires updating several plugins as well, so I'm gonna need to review some breaking changes for those

agilgur5:
Proactively [bumped the repro](https://github.com/agilgur5/rpt2-repro/commit/ebf554eab293ec5dd92b0e01ed36e636834befae) to use Rollup v3 (and TS v5 and rpt2 v0.35)

agilgur5:
Yea I noticed from #447 😅 
ESM adoption has (expectedly) caused a bit of churn in the ecosystem.

Though this PR specifically only needs to update all the Rollup deps, so _maybe_ won't need updating the rest. But seems I might need to do so for #460 anyway

Issue #426, opened 2022-09-26
`0.34.0` not handling JSON import in `rollup.config.ts`


halo951:
## Troubleshooting

1. Does `tsc` have the same output? If so, please explain why this is incorrect behavior

   no

2. Does your Rollup plugin order match [this plugin's compatibility](https://github.com/ezolenko/rollup-plugin-typescript2#compatibility)? If not, please elaborate

 matched 

3. Can you create a [minimal example](https://stackoverflow.com/help/minimal-reproducible-example) that reproduces this behavior? Preferably, use [this environment](https://stackblitz.com/edit/rpt2-repro) for your reproduction
 
https://stackblitz.com/edit/rpt2-repro-h4yxsj?file=rollup.config.ts

## What happens and why it is incorrect
In the `rollup.config.ts`, use the `import pkg from 'package.json'` statement, and the plug-in does not handle. and it is normal in <=v0.33.0

## Environment
```
  System:
    OS: Windows 10 10.0.17763
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 7.05 GB / 15.71 GB
  Binaries:
    Node: 18.9.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.19.1 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    rollup: ^2.79.1 => 2.79.1
    rollup-plugin-typescript2: 0.33.0 => 0.33.0
    typescript: ^4.8.3 => 4.8.3
```
### Versions

v0.34.0

agilgur5:
> and it is normal in <=v0.33.0

Thanks for the repro, I was able to successfully reproduce in `0.34.0` and not in `0.33.0`, getting this error:

```
~/projects/rpt2-repro-h4yxsj 1s
❯ npm run build
$ rollup --config ./rollup.config.ts --configPlugin typescript2
[!] (plugin rpt2) Error: Unexpected token (Note that you need @rollup/plugin-json to import JSON files)
package.json (2:8)
1: {
2:   "name": "rpt2-repro",
           ^
3:   "version": "0.0.0",
4:   "scripts": {

    at error (/home/projects/rpt2-repro-h4yxsj/node_modules/rollup/dist/shared/rollup.js:198:30)
[...]

PR #460, opened 2023-07-18
test: dogfood rpt2 as `configPlugin` w/ `rollup.config.ts`

agilgur5:
### Motivation

- use rpt2 as a `configPlugin` within our own build process, as well as to build itself
  - doubly dogfood

- we don't have any tests for `configPlugin`s, and it's a Rollup CLI option, so this is one of the few ways to test it
  - create a helper `_build-self` script so we don't need to type/repeat as much (or have as much potential for mistakes)
  - this ensures we don't have regressions like #426 
  
### TS Fixes

- replace the unmaintained `rollup-plugin-re` w/ `@rollup/plugin-replace`
  - `rollup-plugin-re` hasn't had a commit in ~5 years and doesn't have typings (natively or in DT)
  - also import the plugins in the same order that they're used

- fix the usage of the `resolve` plugin as it didn't type-check/had outdated usage

### ESM Fixes

- need to use `createRequire` in ESM mode
  - deps: bump `@types/node` to support the types for that as well
  - this is a **BREAKING CHANGE** as it requires Node 14+

- fix `fs-extra` error due to it being CJS

- `require("package.json")` -> `import`
  - and add `resolveJsonModule` to `configPlugin` `tsconfigOverride` to handle this
    - only used by the `configPlugin`, not everything else, so not added to `tsconfig.json`

- output as `build-self/index.mjs` instead of `build-self/dist/rollup-plugin-typescript2.es.js`
  - so that types work out of the box -- due to `index.d.ts`
  - and so that Rollup correctly loads it as ESM and not CJS


agilgur5:
As this is a breaking change due to use of `createRequire`, should probably go out in the same release as #458

agilgur5:
Tests are failing due to use of `createRequire`... this might require bumping `ts-jest` 😕

ezolenko:
I'm also getting this on win, a rollup issue? 
```
> rollup -c rollup.config.self.ts --configPlugin ./build-self/index.mjs="{ tsconfigOverride: { compilerOptions: { resolveJsonModule: true } } }"

[!] Error: Cannot load plugin "C:\sandbox\other\rpt2\rollup-plugin-typescript2\build-self\index.mjs": Only URLs with a scheme in: file and data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'.
Error: Cannot load plugin "C:\sandbox\other\rpt2\rollup-plugin-typescript2\build-self\index.mjs": Only URLs with a scheme in: file and data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'.
    at loadAndRegisterPlugin (C:\sandbox\other\rpt2\rollup-plugin-typescript2\node_modules\rollup\dist\shared\loadConfigFile.js:508:23)
    at addPluginsFromCommandOption (C:\sandbox\other\rpt2\rollup-plugin-typescript2\node_modules\rollup\dist\shared\loadConfigFile.js:458:17)
    at getDefaultFromTranspiledConfigFile (C:\sandbox\other\rpt2\rollup-plugin-typescript2\node_modules\rollup\dist\shared\loadConfigFile.js:583:5)
    at loadConfigFile (C:\sandbox\other\rpt2\rollup-plugin-typescript2\node_modules\rollup\dist\shared\loadConfigFile.js:565:11)
    at Object.loadAndParseConfigFile (C:\sandbox\other\rpt2\rollup-plugin-typescript2\node_modules\rollup\dist\shared\loadConfigFile.js:545:21)
    at getConfigs (C:\sandbox\other\rpt2\rollup-plugin-typescript2\node_modules\rollup\dist\bin\rollup:1691:39)
    at runRollup (C:\sandbox\other\rpt2\rollup-plugin-typescript2\node_modules\rollup\dist\bin\rollup:1665:43)
```

agilgur5:
Dang thought we were past all Windows patching issues 😅

Looks like something is expanding/converting the path in a way Rollup doesn't like `./build-self/index.mjs` -> `C:\sandbox\other\rpt2\rollup-plugin-typescript2\build-self\index.mjs`.
I'm not sure if Rollup is doing that expansion or something else (NPM as it's a `script`?), but I would think Rollup should know how to interpret that correctly 🤔 The error message makes it sound like potentially an ESM spec issue. So if Rollup is the one expanding it, it's expanding incorrectly for ESM. If it's NPM or something else expanding, that's a bit more of a head-scratcher

PR #463, opened 2023-09-26
switching from tslint (deprecated) to eslint

ezolenko:
## Summary

tslint is deprecated and now started to produce errors with new typescript

## Details

Switching to [typescript-eslint](https://typescript-eslint.io/getting-started)


package.json
```diff
@@ -26,17 +26,17 @@
     "build": "rimraf dist/* && rollup -c",
     "watch": "rollup -c rollup.config.self.js -w ",
     "build-self": "rimraf dist/* && rollup -c rollup.config.self.js",
-    "lint": "tslint -c ./tslint.json src/*.ts __tests__/*.ts ./*.js",
+    "lint": "npx eslint .",
```
agilgur5, line 29:
```suggestion
    "lint": "eslint .",
```

`npx` is not needed; `bin` is accessible within `scripts`

.eslintrc.cjs
```diff
@@ -0,0 +1,8 @@
+/* eslint-env node */
+module.exports = {
+	extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended'],
+	parser: '@typescript-eslint/parser',
+	plugins: ['@typescript-eslint'],
+	root: true,
+	ignorePatterns: ['__tests__/**', 'dist/**', 'node_modules/**', 'build-self/**', '*.js'],
```
agilgur5, line 7:
```suggestion
	ignorePatterns: ['dist/**', 'node_modules/**', 'build-self/**', '**/*.js'],
```

I was running `tslint` on tests as well

.eslintrc.cjs
agilgur5, line 1: Can we use `.mjs`? Given that CJS is legacy now

src/index.ts
```diff
@@ -17,8 +17,11 @@ import { TSLIB, TSLIB_VIRTUAL, tslibSource, tslibVersion } from "./tslib";
 import { createFilter } from "./get-options-overrides";
 
 // these use globals during testing and are substituted by rollup-plugin-re during builds
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
agilgur5, line 20: maybe we should just disable this rule entirely. an explicit `any` is _explicit_ after all

agilgur5 (commented):
Quick glance through as I happened to already be on GH and got a notification

.eslintrc.cjs
ezolenko: I think eslint wants cjs, it fails if I just rename the file...

rollup.config.base.js
```diff
@@ -1,3 +1,5 @@
+
+
```
agilgur5, line 2:
```suggestion
```

remove unnecessary new lines

.eslintrc.cjs
```diff
@@ -0,0 +1,11 @@
...
+module.exports = {
+	extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended'],
+	parser: '@typescript-eslint/parser',
+	plugins: ['@typescript-eslint'],
+	root: true,
+	ignorePatterns: ['dist/**', 'node_modules/**', 'build-self/**', '*.js', "__tests__/integration/fixtures/errors/**"],
+	rules: {
+		"@typescript-eslint/no-explicit-any": "off",
+	}
+};
```
agilgur5, line 11:
```suggestion
};

```

newline at EOF

.eslintrc.cjs
agilgur5: huh it does indeed seem to be [unsupported](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-formats). unexpected but I imagine there's a reason for that somewhere

agilgur5 (commented):
looks like CI is passing now -- a few formatting things + a comment, otherwise LGTM

rollup.config.base.js
agilgur5: (also yes, I think we should add `prettier` for formatting, which will be easier now that we no longer have to integrate with deprecated `tslint`)

agilgur5 (approved):
thanks for taking care of this!

rollup.config.base.js
ezolenko: I just tried prettier and it insists on putting opening braces on the same line. >_<

We are long past the time of printing code on transparent slides to project them on a wall trough a contraption of lights and mirrors, so there is no reason for that code style. :)

agilgur5: tbh, I haven't seen your style of opening braces on a new line in like over a decade -- it's very rare 😅 

the main thing with `prettier` and other tools like `standard` (also `isort` et al) is that there is a _consistent_ style, so all code looks the same automatically, with no manual checks needed. (see also [`prettier`'s Philosophy](https://prettier.io/docs/en/option-philosophy), which is very similar to [`standard`'s](https://standardjs.com/#why-should-i-use-javascript-standard-style), albeit sometimes favoring different styles)

braces specifically there is no config option for

Issue #95, opened 2018-06-29
Rollup compilation error when extraneous TS files in project and generating declaration files


amacleay:
#### What happens and why it is wrong

In a project using rollup-plugin-typescript2.

* The project generates a library, so I need it to create type declarations as well as the JS bundle.
* I have typescript files which are not built into the main project.

Compiling results in 

Rolling back https://github.com/ezolenko/rollup-plugin-typescript2/commit/ffe3b6cd9fc03611460aac15d2b880fab7a7d093 fixes this error.

#### Environment 

The following is the set of all files from a reproduction repo:
`$ for f in $(git ls-files); do echo -e "\n// $f"; cat $f; done`

#### Versions
- typescript: 2.9.2
- rollup: 0.62.0
- rollup-plugin-typescript2: 0.15.1

#### rollup.config.js

```js
// rollup.config.js  
import typescript from 'rollup-plugin-typescript2';
  
export default {   
  input: 'index.ts',   
  output: [{ file: 'dist/index.js', format: 'es' }],    
  plugins: [    
    typescript({ verbosity: 3 }),     
  ],   
};   
```

#### tsconfig.json

```
// tsconfig.json     
{    
  "compilerOptions": {
    "declaration": true
  }    
}
```

#### package.json

```json
// package.json
{    
  "name": "@cht/demo-rollup-issue",
  "private": true,     
  "engines": {
    "node": ">=8"     
  },
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"     
  },    
  "devDependencies": { 
    "rollup": "0.62.0", 
    "rollup-plugin-typescript2": "0.15.1",   
    "ts-node": "7.0.0",    
    "typescript": "2.9.2"
  }   
}
```

#### the rest of the code

```typescript   
// extra/extra.ts
console.log('read all about it');   
```

```typescript
// index.ts   
module.exports = 1;
```

#### plugin output with verbosity 3

```

index.ts → dist/index.js...
rpt2: typescript version: 2.9.2
rpt2: tslib version: 1.9.2
rpt2: rollup-plugin-typescript2 version: 0.15.1
rpt2: plugin options:
{
    "verbosity": 3,
    "check": true,
    "clean": false,
    "cacheRoot": "/tmp/cht-demo-rollup-issue/.rpt2_cache",
    "include": [
        "*.ts+(|x)",
        "**/*.ts+(|x)"
    ],
    "exclude": [
        "*.d.ts",
[...]

PR #476, opened 2025-05-07
fix(host): invalidate cache less to speed up compilation for large projects

std4453:
## Summary

Fixes a few problems related to `host.ts` and speed up compilation for large projects.

## Details

The current version of `rollup-plugin-typescript2` causes the TypeScript cache to get constantly invalidated, for each entry file during a single rollup build, causing build time to increase non-linearly. For for fairly complicated projects, this easily bloat into minutes (about 400s for our project with ~50 files and ~15000 locs).

I've identified and fixed these problems that caused this:

- In `setSnapshot`, file version is increased unnecessarily when source hasn't changed.
- Files in `node_modules` are added incorrectly into LanguageServiceHost.fileNames by `setSnapshot`, since the function is called indirectly for all dependency files discovered while type checking. `fileNames` should be the list of entry files for this build, there's no point updating it during the build process.
- In `getScriptSnapshot`, also create snapshot when file is empty, otherwise it would be marked as 'missing' and invalidate the cache.

I've verified that the updated code can compile itself and our company project, making absolutely no changes to the output. The new compile time is 14s compared to 400s.

You can override `rollup-plugin-typescript2` with `@std4453/rollup-plugin-typescript2` to try it out, which is forked from version `0.36.1`.

### DIFF
--- a/src/host.ts
+++ b/src/host.ts
@@ -11,7 +11,7 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
 	private service?: tsTypes.LanguageService;
 	private fileNames: Set<string>;
 
-	constructor(private parsedConfig: tsTypes.ParsedCommandLine, private transformers: TransformerFactoryCreator[], private cwd: string)
+	constructor(private parsedConfig: tsTypes.ParsedCommandLine, private transformers: TransformerFactoryCreator[], private cwd: string, private filter: (id: string | unknown) => boolean)
 	{
 		this.fileNames = new Set(parsedConfig.fileNames);
 	}
@@ -39,6 +39,9 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
 		const snapshot = tsModule.ScriptSnapshot.fromString(source);
 		this.snapshots[fileName] = snapshot;
 		this.versions[fileName] = (this.versions[fileName] || 0) + 1;
+
+		if (this.filter(fileName))
+			this.fileNames.add(fileName);
 
 		return snapshot;
 	}
--- a/src/host.ts
+++ b/src/host.ts
@@ -31,11 +31,10 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
 	{
 		fileName = normalize(fileName);
 
-		const originalSnapshot = this.snapshots[fileName];
-
-		if (originalSnapshot && originalSnapshot.getText(0, originalSnapshot.getLength()) === source) {
-			return originalSnapshot;
-		}
+		// don't update the snapshot if there are no changes
+		const prevSnapshot = this.snapshots[fileName];
+		if (prevSnapshot?.getText(0, prevSnapshot.getLength()) === source)
+			return prevSnapshot;
 
 		const snapshot = tsModule.ScriptSnapshot.fromString(source);
 		this.snapshots[fileName] = snapshot;
--- a/src/host.ts
+++ b/src/host.ts
@@ -31,10 +31,16 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
 	{
 		fileName = normalize(fileName);
 
+		const originalSnapshot = this.snapshots[fileName];
+
+		if (originalSnapshot && originalSnapshot.getText(0, originalSnapshot.getLength()) === source) {
+			return originalSnapshot;
+		}
+
 		const snapshot = tsModule.ScriptSnapshot.fromString(source);
 		this.snapshots[fileName] = snapshot;
 		this.versions[fileName] = (this.versions[fileName] || 0) + 1;
-		this.fileNames.add(fileName);
+
 		return snapshot;
 	}
 
@@ -46,7 +52,7 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
 			return this.snapshots[fileName];
 
 		const source = tsModule.sys.readFile(fileName);
-		if (source)
+		if (source !== undefined)
 			return this.setSnapshot(fileName, source);
 
 		return undefined;
--- a/src/host.ts
+++ b/src/host.ts
@@ -27,14 +27,19 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
 		this.service = service;
 	}
 
-	public setSnapshot(fileName: string, source: string): tsTypes.IScriptSnapshot
+	public setSnapshot(fileName: string, source: string, forcedUpdate = false): tsTypes.IScriptSnapshot
 	{
 		fileName = normalize(fileName);
 
 		// don't update the snapshot if there are no changes
 		const prevSnapshot = this.snapshots[fileName];
-		if (prevSnapshot?.getText(0, prevSnapshot.getLength()) === source)
+		if (prevSnapshot?.getText(0, prevSnapshot.getLength()) === source) {
+			if (forcedUpdate) {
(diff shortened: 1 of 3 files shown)


std4453:
@agilgur5 I think that CI is failing because the latest version of `actions/setup-node` is now `v4`.

ezolenko:
@std4453 yeah, I fixed it in master yesterday, if you merge CI should work

agilgur5:
Yes you should be able to rebase on top of #477 to get CI to run properly.

I haven't fully reviewed this, but it might be good to add regression tests for these changes.

src/host.ts
```diff
@@ -31,10 +31,16 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
...
+
+		if (originalSnapshot && originalSnapshot.getText(0, originalSnapshot.getLength()) === source) {
+			return originalSnapshot;
+		}
+
 		const snapshot = tsModule.ScriptSnapshot.fromString(source);
 		this.snapshots[fileName] = snapshot;
 		this.versions[fileName] = (this.versions[fileName] || 0) + 1;
-		this.fileNames.add(fileName);
+
```
agilgur5, line 43: > Files in `node_modules` are added incorrectly into LanguageServiceHost.fileNames by `setSnapshot`, since the function is called indirectly for all dependency files discovered while type checking. `fileNames` should be the list of entry files for this build, there's no point updating it during the build process.

Are you sure it's only for entry files? Do you have a specific source for this?
The TS Compiler wiki example [here](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#incremental-build-support-using-the-language-services) does use the variable `rootFileNames` in a way that sounds like it means only entry files.

But per https://github.com/ezolenko/rollup-plugin-typescript2/pull/364 there were some bugs with this before when the Set was incomplete.
In [the diff in the commit prior to that PR](https://github.com/ezolenko/rollup-plugin-typescript2/commit/c86e07bc0f2f875e3953ee651871830351d38452?w=1#diff-49ecc36f34e1f45bd5ed5039646b8eaa1e919ecf4bcdfed5eae5c356bf11bff8R39), the `snapshots` dictionary (object) was used before the Set was created, so it always contained all the files in rpt2's implementation (it may have been incorrect since inception though)

agilgur5 (commented):
My memory is pretty foggy with this codebase right now, especially with regard to [the sparsely documented TS Compiler internals](https://github.com/ezolenko/rollup-plugin-typescript2/pulls?q=is%3Apr+is%3Aopen+label%3A%22topic%3A+TS+Compiler+API+Docs%22), but I left a few comments in-line

src/host.ts
```diff
@@ -46,7 +52,7 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
 			return this.snapshots[fileName];
 
 		const source = tsModule.sys.readFile(fileName);
-		if (source)
+		if (source !== undefined)
```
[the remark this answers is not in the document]
std4453: I found a found in one of our dependencies which is simply an empty file, and it became the reason why the TypeScript cache was marked dirty for every entry file. Seeing that the host didn't return an `snapshot`, TypeScript marked the file as 'missing', so maybe it was deleted, so it would have to rebuild cache, which makes sense to me.

std4453:
@ezolenko I've rebased my branch to include the CI fix, look like it needs approval from maintainers to run.

src/host.ts
```diff
@@ -31,10 +31,16 @@ export class LanguageServiceHost implements tsTypes.LanguageServiceHost
 	{
 		fileName = normalize(fileName);
 
+		const originalSnapshot = this.snapshots[fileName];
+
+		if (originalSnapshot && originalSnapshot.getText(0, originalSnapshot.getLength()) === source) {
+			return originalSnapshot;
+		}
```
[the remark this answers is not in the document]
std4453: What TypeScript does is that it first calls `getScriptVersion` to check if anything has changed, then `getScriptSnapshot` to get the new snapshot. In a typical implementation, change of version should happen **before** building new snapshot. It should actually happen before calling `getEmitOutput`, which exactly what the official example does.

Here I honestly don't understand why version is updated inside `getSnapshot`, however in order to not break anything, I compared the current snapshot to the requested source, which should be okay because `ScriptSnapshot.fromString` is simply creating a `StringScriptSnapshot` object, which holds no state other than `source`, and never writes to it, so it should behave identically in every way to the previous snapshot.

In a typical implementation where hosts listen to FS events for file updates, there's usually only one changed file, however that information isn't available to TypeScript, so it needs to check every imported file for changes. Reading files from slow filesystems might take hundreds of milliseconds, so TypeScript opens up an optimizing opportunity to the host by separating change check into two phases, first `getScriptVersion`, then `getScriptSnapshot`.

src/host.ts
std4453: This seems to be a bit complicated. Here's my analysis, please point me out if I was wrong.

The issues leading to adding files provided to `setSnapshot` are both related to TypeScript's `getValidSourceFile`, which is only called for emit and type-check roots, so TypeScript does not allow missing files (instead of imported files, which TypeScript simply marks their types as any without throwing an error).

The fix for #95 works because it added all files from `parsedConfig` into `getScriptFileNames`, which include extra files covered by `include` from `tsconfig.json`. rpt2 tries to generate missed declarations by transforming all files from `parsedConfig`, which calls `getEmitOutput` and indirectly `getValidSourceFile`, so it makes sense to include these into `getScriptFileNames`.

At that point, the fix resulted in imported files being excluded from `getScriptFileNames`, since they're discovered by `getScriptSnapshot` and not added to `this.fileNames`. Meanwhile, at build end under watch build, rpt2 type-checks all files on import tree, including those that are out of the root directory and therefore not covered by `parsedConfig`, causing problems like #364.

However, during the build of import tree, the modules are filtered against the `include` and `exclude` from `parsedConfig`, so imports from `node_modules` are not included and not type-checked. Nevertheless, `setSnapshot` **does not** check whether files should be type-checked, before inserting them into `this.fileNames`.

 So in my case, I found that after building the first file, `fileNames` got stuffed with files from `node_modules` (which are excluded by tsconfig), although this would be unnecessary, since rpt2 would never call `transform` on these files.

I think the correct solution is to reuse the filter from `createFilter` and filter files before adding them to `fileNames`, perhaps padding `filter` down to constructor of `LanguageServiceHost`?

agilgur5:
I approved it before (when it failed due to old `actions/setup-node@v2`), but apparently I need to do so again after the rebase.
It has ran now and errors out as expected on the host tests, which would need adjusting.

src/host.ts
std4453: I've passed `filter` down to `LanguageServiceHost` and restored adding to `fileNames`.

std4453:
OK I think I've resolved code comments, however I couldn't get the watch CI spec working, even locally. Weirdly though, even if I revert the code I've edited, the test would still fail. Any idea on why that might happen? @agilgur5

agilgur5:
Hey @std4453, thanks for the iterations and comments!

Just wanted to leave a heads-up that my family and I have COVID right now (again), so it's gonna be a bit before I can review this in-depth as my brain is not running too well right now 🤕😷

ezolenko:
@agilgur5 this looks good to me, do you have any problems with merging it and releasing as 0.38?

